PR #132, opened 2025-04-11
Add PHQ-9 followup extension

aduane:
This shows an example of originating a questionnaire based on the response to one that was just committed.
In this case, a PHQ-2 with a score above 2 will cause a PHQ-9 to be originated, and the questions that are common between the two questionnaires will be pulled forward to prevent the need for double-entry.

https://github.com/user-attachments/assets/2808f05b-7c99-4a18-86a2-279bd38ee600

### DIFF
--- a/extensions/phq9_followup/handlers/phq9_followup.py
+++ b/extensions/phq9_followup/handlers/phq9_followup.py
@@ -1,0 +1,86 @@
+from uuid import uuid4
+
+from canvas_sdk.commands import QuestionnaireCommand
+from canvas_sdk.effects import Effect
+from canvas_sdk.events import EventType
+from canvas_sdk.handlers import BaseHandler
+
+from canvas_sdk.v1.data.command import Command
+from canvas_sdk.v1.data.questionnaire import Questionnaire, Interview, InterviewQuestionResponse
+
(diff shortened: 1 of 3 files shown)


extensions/phq9_followup/handlers/phq9_followup.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Check the score of this PHQ-2
+        phq_2_interview = event_command.anchor_object
+        score = self.score(phq_2_interview)
+        if score and score <= 2:
+            # If the score is not abnormal, no PHQ-9 needed
+            return []
+
+        # Prep a new questionnaire command, selecting the PHQ-9
+        phq_9 = Questionnaire.objects.filter(code_system=PHQ_CODE_SYSTEM, code=PHQ9_CODE).first()
```
rmagier1, line 46: You might want to filter by `can_originate_in_charting=True` to make sure you are getting the current version that frontend would be using

aduane: @rmagier1 fixed in [6f4d548](https://github.com/Medical-Software-Foundation/canvas/pull/132/commits/6f4d54854ea2c9fac5b8f0adb121136cb790a682)

PR #141, opened 2025-05-10
Add visualizations for vitals and questionnaire scoring trends to see in patient summary

rmagier1:


extensions/visualizations/scoring_visualizer/CANVAS_MANIFEST.json
```diff
@@ -0,0 +1,38 @@
+{
+    "sdk_version": "0.35.0",
```
mcb041, line 2: @rmagier1 Is this the right SDK version? shouldn't it be 0.1.4?

PR #202, opened 2025-12-16
Update apppointments.sql to use SDK views

aduane:
By tying as much external access to our database to the Canvas SDK as possible, we can minimize the risk of introducing breaking changes. Canvas engineers will change internal schema and only ensure internal usages do not break. Canvas engineers will ensure backwards compatibility in the SDK views, however. It is required these views remain backwards compatible so deployed plugins do not break.

Building reports on the SDK database views provides more guarantees against breakage, and the documentation for the data present in these views is accessible on the docs site. 

Example: https://docs.canvasmedical.com/sdk/data-appointment/#appointment

The attribute names documented in the data module are not 1:1 to the database views, but are close enough to represent a marked improvement over the out-of-data generated schema docs we've used historically.

### DIFF
--- a/sample-sql/apppointments.sql
+++ b/sample-sql/apppointments.sql
@@ -2,24 +2,21 @@
 
 SELECT
     -- Appointment details
-    apt.externally_exposable_id AS apt_UUID,
-    an.id AS appt_note_id,
+    apt.id AS apt_UUID,
+    an.dbid AS appt_note_id,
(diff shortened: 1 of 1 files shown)


rmagier1:
@JessicaHerbert What are we doing with this PR, I know you manage these sql queries

JessicaHerbert:
we can close this - thanks for the example AD

PR #217, opened 2026-01-28
Note Creation API

rmagier1:


extensions/note-command-api/note_command_api/protocols/note_api.py
```diff
@@ -1,12 +1,198 @@
...
+            return [
+                JSONResponse(
+                    {"errors": errors},
+                    status_code=HTTPStatus.BAD_REQUEST,
+                )
+            ]
+
+        note_effect = NoteEffect(
+            instance_id=instance_id,
+            note_type_id=note_type_identifier,
```
jrwils: It looks like you are using note_type_id correctly, but for some reason the note header is showing as None when I test this:

Image

jrwils (approved):

PR #228, opened 2026-04-03
Fix encounter_list memory by using DB-level pagination

alexiadowns-canvas:
## Summary
- **Both pagination paths** (`_sort_and_paginate_database` and `_sort_and_paginate_delegated_orders`) were calling `list(note_queryset)`, loading the **entire result set into memory** before slicing for pagination. For large instances like Doctronic, this caused **100–500 MB memory spikes per request**, leading to OOM-related 502 outages.
- **Normal sort path**: replaced `list(queryset)` + Python slicing with Django queryset slicing (`queryset[offset:offset+page_size]`), which translates to SQL `OFFSET`/`LIMIT`.
- **Delegated orders path**: replaced the full-queryset materialization + per-note Python loop (N+1 queries) with DB-level `Subquery` annotations on `ImagingOrder.delegated` and `Referral.forwarded`, enabling database-side sort, filter, and pagination.
- The exact delegated orders count (which checks open tasks via the JSON `task_ids` field) is still computed per-note in the render loop — but only for the ~25 notes on the displayed page, not all notes in the system.
- Removed the now-unused `_apply_pagination` helper.

## Test plan
- [ ] Deploy to a staging instance and verify encounter list loads correctly with default sort (date of service)
- [ ] Verify sorting by each column still works (patient name, provider, location, note title, DOS, billable, uncommitted commands, delegated orders, claim queue)
- [ ] Verify the "has delegated orders" filter returns correct results
- [ ] Verify pagination (next/previous, page numbers) works correctly
- [ ] Monitor plugin runner memory logs after deploy — spikes should drop from 100-500MB to single-digit MB

### DIFF
--- a/extensions/encounter_list/applications/my_application.py
+++ b/extensions/encounter_list/applications/my_application.py
@@ -1,7 +1,8 @@
 from http import HTTPStatus
 
 import arrow
-from django.db.models import Count, Q, F, Case, When, Value
+from django.db.models import Count, Q, F, Case, When, Value, Subquery, OuterRef, IntegerField
+from django.db.models.functions import Coalesce
 
 from canvas_sdk.effects import Effect
 from canvas_sdk.effects.launch_modal import LaunchModalEffect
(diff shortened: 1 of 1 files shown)


extensions/encounter_list/applications/my_application.py
```diff
@@ -236,31 +237,45 @@ def _get_sort_fields(self, sort_by: str) -> list[str]:
...
-                continue
-            notes_with_delegated_count.append((note, delegated_commands))
-        
-        # Sort by delegated orders count
-        notes_with_delegated_count.sort(
-            key=lambda x: x[1], 
-            reverse=(sort_direction == "desc")
+        """Sort and paginate for delegated orders using database-level annotation.
+
+        Uses ImagingOrder and Referral FKs to Note to compute an approximate
```
djantzen, line 242: I gather that this change will result in a higher count on some views than on others, because it will be the superset of all tasks, not just those that are open. Is that change in behavior going to be acceptable to users?

extensions/encounter_list/applications/my_application.py
```diff
@@ -236,31 +237,45 @@ def _get_sort_fields(self, sort_by: str) -> list[str]:
...
-        
-        # Apply pagination
-        return self._apply_pagination(sorted_notes, page, page_size)
+        forwarded_referral_count = (
+            Referral.objects.filter(note=OuterRef("pk"), forwarded=True)
+            .values("note")
+            .annotate(cnt=Count("id"))
+            .values("cnt")
+        )
+        note_queryset = note_queryset.annotate(
```
djantzen, line 259: We are missing indexes on `forwarded` and `delegated`, which might force a sequential scan. We need to look at an `explain` plan for this new query with the annotation and subqueries.

djantzen: @claude What SQL will be generated by lines 259-264?

PR #240, opened 2026-04-28
Fix ChartEventPublisher target ID extraction

aduane:
## Summary
- `self.event.target` is a `TargetType` dataclass with `.id`/`.type` attributes, not a UUID string. The handler did `str(target_id)` which produced the dataclass repr (e.g. `\"TargetType(id='ccc0a14e-...', type=<class '...Interview'>)\"`) and then passed it to `model.objects.filter(id=...)`, raising `badly formed hexadecimal UUID string` for every event in `EVENT_MAP`. The chart-event broadcaster has been silently failing for every chart change since this code shipped — no patient channel ever got a section-refresh ping.
- Extract `target.id` from the dataclass; preserve the string-target path the existing tests use and the `context["target"]` fallback.
- Added a regression test covering the dataclass-shaped target.

## Test plan
- [x] `pytest extensions/provider_clinical_summary_companion/tests/handlers/test_chart_event_publisher.py` — 11 passed
- [x] After deploy, edit a condition/medication/allergy/immunization/interview/vital and confirm the patient channel receives a section-refresh broadcast (no more UUID validation traceback in plugin-runner logs)

### DIFF
--- a/extensions/provider_clinical_summary_companion/provider_clinical_summary_companion/handlers/chart_event_publisher.py
+++ b/extensions/provider_clinical_summary_companion/provider_clinical_summary_companion/handlers/chart_event_publisher.py
@@ -59,10 +59,11 @@ class ChartEventPublisher(BaseHandler):
             return []
         model, sections = mapping
 
-        target_id = getattr(self.event, "target", None)
-        if target_id is None:
+        target = getattr(self.event, "target", None)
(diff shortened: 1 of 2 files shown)


aduane:
Confirmed this works in my local env

aduane:
Also updated in training

PR #247, opened 2026-05-06
create candid integration plugin

mbiannaccone:


mbiannaccone:
@claude review

extensions/candid/candid/api/claim_detail.py
```diff
@@ -0,0 +1,163 @@
...
+    return info
+
+
+class CandidClaimDetailAPI(SimpleAPIRoute):
+    """Serve Candid timeline data and trigger syncs for a single claim."""
+
+    PATH = "/claim-detail"
+
+    def authenticate(self, credentials: SessionCredentials) -> bool:
+        return bool(credentials.logged_in_user.get("id"))
```
rmagier1: Can we just use the https://docs.canvasmedical.com/sdk/handlers-simple-api-http/#staff-session

extensions/candid/candid/api/client.py
```diff
@@ -0,0 +1,249 @@
...
+        """Update an existing encounter via PATCH /encounters/v4/{encounter_id}.
+
+        Used when re-submitting a claim that was already filed — the encounter
+        exists on Candid's side and needs its data updated rather than a new
+        encounter created.
+
+        Returns ``(was_successful, message)``. On success, ``message`` is the
+        Candid encounter ID. On failure, ``message`` is a human-readable error.
+        """
+        response = requests.patch(
```
rmagier1: We deprecated the use of the `requests` library in favor of https://docs.canvasmedical.com/sdk/utils/#making-requests-with-http

extensions/candid/candid/api/dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+PAGE_SIZE = 50
+
+
+class CandidDashboardAPI(SimpleAPIRoute):
+    """List Candid-submitted claims with status for the dashboard application."""
+
+    PATH = "/dashboard"
+
+    def authenticate(self, credentials: SessionCredentials) -> bool:
+        return bool(credentials.logged_in_user.get("id"))
```
rmagier1: Again should we use the staff mixin credentials?

extensions/candid/candid/api/dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+        else None
+    )
+
+    patient = getattr(claim, "patient", None)
+    if patient:
+        first = (patient.first_name or "").strip()
+        last = (patient.last_name or "").strip()
+        patient_name = f"{first} {last}".strip() or "(unnamed)"
+        patient_id = str(patient.id)
+    else:
```
rmagier1, line 149: Can a claim really have no patient? Or like candid sent us information that we can't match to a patient in canvas?

extensions/candid/candid/applications/candid_dashboard.py
```diff
@@ -0,0 +1,722 @@
...
+
+    def on_open(self) -> Effect:
+        return LaunchModalEffect(
+            content=_html(),
+            target=LaunchModalEffect.TargetType.PAGE,
+            title="Candid Dashboard",
+        ).apply()
+
+
+def _html() -> str:
```
rmagier1: All html should be in a .html file (also css should be in a .css file)

extensions/candid/candid/models/sync_state.py
```diff
@@ -0,0 +1,24 @@
...
+from canvas_sdk.v1.data.base import CustomModel
+
+LOG_TYPE_SYNC = "sync"
+LOG_TYPE_PAYMENT_REPORTED = "payment_reported"
+
+
+class SyncLog(CustomModel):
+    """One row per activity event per claim."""
+
+    canvas_claim_id = TextField()
```
rmagier1: Can this be a Foreign Key to the claim instead

extensions/candid/candid/api/claim_detail.py
mbiannaccone: i remember trying to use that at the very beginning and it wasn't working. i will try again and test against training!

extensions/candid/candid/api/client.py
mbiannaccone: good call! will do

extensions/candid/candid/api/dashboard.py
mbiannaccone: technically, yes it is possible (ClaimPatient is a 1:1 fk to Claim). it is very very unlikely, but imo worth the defensive code just in case a ClaimPatient row somehow gets deleted , or not created.

extensions/candid/candid/applications/candid_dashboard.py
mbiannaccone: good call!

extensions/candid/candid/models/sync_state.py
mbiannaccone: ya we could do that eventually. i'm also considering moving this to claimmetadata (where other candid data is stored) just to be consistent.

mbiannaccone: migrated synclog to claimmetadata (which runs at 2am installation timezone during the nightly sync cron). tested it on training last night and it worked perfectly! going to install to the 5 instances today.

mbiannaccone: actually found a small issue - claims that also got a new sync after the migration lost their historical syncs (they are only showing the most recent sync). updating so that the migration runs on its own cron 2 hours before the sync cron.

mbiannaccone: plugin installed on all 5 instances. at midnight tonight, all synclog rows should be migrated to claimmetadata. i have investigator monitoring for me!

mbiannaccone:
synclog migration to claimmetadata was successful on all 5 instances last night. uploading newest version shortly

rmagier1 (approved):

PR #251, opened 2026-05-07
Update last reviewed section

animassaro:
## Summary

  Iterative work on top of `last_reviewed` plugin. Two themes:                                                              
  
  - **Overflow affordance.** When the section's six rows extend below the host's chart-summary slot, a soft fade now appears at the bottom edge as a "more below" hint. A sticky `::after` overlay paints the fade so it always sits at the visible viewport bottom (not the bottom of all content), and a small client-side script (`static/section.js`) walks up from `.lr-section` to find the nearest scrolling ancestor, listens for scroll/resize, and toggles a `.lr-section--at-bottom` modifier class so the fade drops out once the user has scrolled to the section's bottom edge. The handler injects the script via `render_to_string` exactly the way it already injects the CSS, since the chart-summary content ships as one inline blob.
  - **README expansion.** Adds *Problem this solves*, *Who it's for*, and *How to install* sections so a reader landing on the plugin can tell at a glance why it exists, who benefits, and that no secrets/env vars are required (only the section-ordering caveat, already documented). 

  - The branch is also where we *learned* a few things about the chart-summary render context — captured in commit messages so future readers know why the code looks the way it does:                                                                                                                                                    
  - The host's wrapper owns the scroll, not us. An earlier attempt to be the scroll container ourselves produced nested scrollbars (commit `653c759` reverts that).
  - Native macOS Chrome/Safari respect the OS "show scrollbars when scrolling" preference even with explicit `::-webkit-scrollbar` styling, so a forced-persistent scrollbar isn't reachable from CSS alone — hence the fade affordance instead.                                                      
  - The host's render context doesn't honor an `opacity` transition on the `::after` pseudo, so the fade snaps in/out rather than animating; the unused `transition` rule was trimmed.                                                                                                                      
                  
  ## Test plan                                                                                                                                         
                  
  - [ ] Open a patient chart with all six chart sections populated; confirm the *Last Reviewed* section still renders correctly above the standard     
  sections.
  - [ ] Resize the browser or navigate to a chart layout where the section overflows the host's slot; confirm a soft fade appears at the bottom edge of
   the visible section.                                                                                                                                
  - [ ] Scroll down within the chart-summary column until the section's last row is fully in view; confirm the fade has disappeared.
  - [ ] Scroll back up; confirm the fade reappears.                                                                                                    
  - [ ] On a layout where the section fits without overflow, confirm there's nothing visually awkward at the bottom edge.
  - [ ] Confirm the rollback-on-deletion behavior from PR #1 still works (mark a section reviewed → refresh → delete the review → refresh → row        
  reverts).                                                                                                                                            
                                                                                                                                                       
  ## Files changed                                                                                                                                     
                  
  - `last_reviewed/static/section.js` (new) — scroll-aware fade visibility                                                                             
  - `last_reviewed/static/section.css` — sticky `::after` overlay + `.lr-section--at-bottom` rule
  - `last_reviewed/static/section.html` — adds the `<script>` tag                                                                                      
  - `last_reviewed/handlers/section_content.py` — loads + injects `section.js`
  - `last_reviewed/README.md` — adds problem/audience/install sections, documents the fade affordance                                                  
  - `last_reviewed/CANVAS_MANIFEST.json` — `0.1.8` → `0.1.18`           -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -   -

### DIFF
--- a/extensions/last_reviewed/last_reviewed/static/section.css
+++ b/extensions/last_reviewed/last_reviewed/static/section.css
@@ -6,15 +6,23 @@
   box-sizing: border-box;
   max-height: 280px;
   overflow-y: auto;
-  /* Static "there's more below" hint: an inset shadow at the bottom edge
-     of the visible area. Doesn't depend on the host's background color
-     and never has to fight macOS's hide-when-idle scrollbar behavior --
-     the native scrollbar still works during scroll, this just gives an
-     idle-time affordance that the section is scrollable. */
-  box-shadow: inset 0 -16px 16px -16px rgba(0, 0, 0, 0.18);
   /* Firefox: still ask for the thin native scrollbar where supported. */
   scrollbar-width: thin;
   scrollbar-color: rgba(0, 0, 0, 0.3) transparent;
+}
+/* Always-visible "more below" overlay. Lives on top of the content (an
+   inset box-shadow is painted behind content and was only visible at
+   the very bottom of scroll). Sticky to bottom: 0 keeps it at the
+   bottom edge of the visible scroll viewport at every scroll position. */
+.lr-section::after {
+  content: "";
+  display: block;
+  position: sticky;
+  bottom: 0;
+  height: 36px;
+  margin-top: -36px;
+  pointer-events: none;
+  background: linear-gradient(to bottom, transparent, rgba(0, 0, 0, 0.28));
 }
 .lr-section::-webkit-scrollbar {
   width: 8px;
--- a/extensions/last_reviewed/last_reviewed/static/section.css
+++ b/extensions/last_reviewed/last_reviewed/static/section.css
@@ -10,10 +10,11 @@
   display: block;
   position: sticky;
   bottom: 0;
-  height: 36px;
-  margin-top: -36px;
+  height: 14px;
+  margin-top: -14px;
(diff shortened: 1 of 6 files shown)


aduane (approved):

extensions/last_reviewed/last_reviewed/static/section.css
```diff
@@ -5,6 +5,24 @@
...
 }
+.lr-section::after {
+  content: "";
+  display: block;
+  position: sticky;
+  bottom: 0;
+  height: 36px;
+  margin-top: -36px;
+  pointer-events: none;
+  background: linear-gradient(to bottom, transparent, rgba(0, 0, 0, 0.22));
```
aduane: The fade is certainly there, but with a few changes it can be a little "softer". This is probably subjective.

before:
Image

after:
Image


```suggestion
  height: 14px;
  margin-top: -14px;
  border-radius: 14px;
  pointer-events: none;
  background: linear-gradient(to bottom, transparent, rgba(0, 0, 0, 0.1));
```

aduane (approved):

PR #264, opened 2026-05-13
Add intake-assignment-panel plugin

amandacanvas:
## Summary

Adds **intake-assignment-panel**, a provider-menu Canvas application that surfaces pending intake notes filtered by the logged-in clinician's active state licenses.

In multi-state telehealth practices, a queue of unsigned intake notes typically lands in front of every clinician at once — even though each clinician can only take patients in the specific states where they're licensed. Clinicians end up scanning the full queue and skipping past work they can't legally pick up, which slows triage and leaves intakes lingering. This plugin filters the queue server-side using each clinician's `StaffLicense` records, so the panel only shows intakes the clinician is licensed to handle.

## What's included

- **Application** (`IntakePanelApp`) — provider-menu item that launches the panel in a modal.
- **SimpleAPI** (`IntakeAPI`) — `StaffSessionAuthMixin`-gated endpoints that render an HTMX-driven sortable table.
- **Filtering logic** — queries active (non-expired) `StaffLicense` records of type `STATE_LICENSE` for the logged-in user; matches against the patient's home-address `state_code`.
- **Fallback behavior** — if the logged-in user has zero active state licenses on file, the panel shows all pending intakes unfiltered (so admin/triage staff aren't accidentally locked out).
- **Configurable note types** — `INTAKE_NOTE_TYPES` secret accepts a comma-separated list of Note Type names to treat as "intake".
- **Tests** — 38 unit tests covering filter, sort, humanize-time helpers, the ORM-backed row builder, and the SimpleAPI endpoints. 100% line + branch coverage.

## Configuration

| Secret | Required | Purpose |
|---|---|---|
| `INTAKE_NOTE_TYPES` | Yes | Comma-separated list of Note Type names to treat as intake (case-insensitive). |
| `CANVAS_INSTANCE_URL` | Yes | Canvas instance base URL, used to construct deep-links to each unsigned note. |

## Author

Amanda Peterson <amanda.peterson@canvasmedical.com>

## Test plan

- [ ] `uv sync` resolves cleanly from the included `uv.lock`.
- [ ] `uv run pytest tests/` passes (38 tests).
- [ ] `uv run pytest --cov=intake_assignment_panel --cov-branch` shows 100% coverage.
- [ ] After `canvas install intake-assignment-panel` and setting both secrets, the "Intake Assignment Panel" menu item appears under the provider menu.
- [ ] Opening the panel as a staff member with active state licenses on their profile shows a filter banner listing those states and only intakes whose patient's home state matches.
- [ ] Opening the panel as a staff member with no active state licenses shows the unfiltered banner and all pending intakes.
- [ ] Clicking a row opens the unsigned note in the patient chart in a new tab.
- [ ] Sort headers (Patient, State, Time pending) cycle asc → desc → default; default sort is oldest pending first.


amandacanvas:
@claude review once

PR #274, opened 2026-05-15
Add expired-icd10-alert plugin

csande:
## Summary

Open-sources the `expired-icd10-alert` plugin from `gtm-extensions`. The plugin annotates patient-chart conditions with an **EXPIRED** tag when their ICD-10 code has been deleted from the current CMS code set, helping providers identify conditions that need to be re-coded.

The source plugin had no organization-specific code, so this is primarily a restructuring + maintainability pass.

## Improvements over the source plugin

- **Expired-codes list moved out of Python.** Default list now lives in `expired_icd10_alert/data/expired_icd10_codes.json` with `code_set_version`, `effective_date`, and CMS source URL. CMS updates become a reviewable JSON diff instead of a Python edit.
- **Optional `EXPIRED_ICD10_CODES_OVERRIDE` secret.** When set, replaces the bundled list at runtime — lets a site patch the list (e.g. mid-cycle errata, locally deprecated codes) without waiting for a plugin release.
- **Accepts both ICD-10 system strings.** `"ICD-10"` and the FHIR URI `"http://hl7.org/fhir/sid/icd-10"`.
- **Skips `entered_in_error` conditions.**
- **Modern `BaseHandler`** from `canvas_sdk.handlers` (was `BaseProtocol`) and `self.event.context` (was `self.context`).
- **Repo conventions:** minimal `pyproject.toml` (Python ≥3.12, `canvas[test-utils]` only), `mypy.ini`, MIT `LICENSE` file, `sdk_version` `0.94.0`.
- **README rewritten** to document the data file as the single source of truth and the override-secret format; per-code list removed from the README to prevent drift with the JSON.
- **Tests tightened**: dropped the trivial "iterate every code in the set" test; added override, FHIR-URI, and `entered_in_error` cases. 22 tests, 100% line coverage, mypy clean.

## Test plan

- [x] `uv run pytest -v --cov=expired_icd10_alert` — 22 passed, 100% coverage
- [x] `uv run mypy expired_icd10_alert/` — no issues
- [ ] `uv run canvas install .` on a Canvas instance
- [ ] Add a `Diagnose` for `G35` (Multiple Sclerosis) → condition shows the EXPIRED tag
- [ ] Add a `Diagnose` for `I10` (Essential Hypertension) → no tag
- [ ] Set `EXPIRED_ICD10_CODES_OVERRIDE = "I10"`, reload → `I10` now tags EXPIRED; `G35` does not
- [ ] Add a `Diagnose` for `S30.1XXA` → tag appears (encounter-type suffix path)


extensions/expired-icd10-alert/expired_icd10_alert/protocols/expired_icd10_annotation.py
aduane, line 1: rename the `protocols` directory to `handlers`

PR #276, opened 2026-05-15
Add provider-availability-manager plugin

amandacanvas:
## Summary

A drag-and-schedule weekly calendar for defining provider availability, paired with a slot-filter protocol that enforces those windows inside Canvas's native appointment book.

- **Manage Availability UI** — Open from the three-line/hamburger menu. Staff drag to create Available (Clinic) and Busy (Administrative) windows for any provider at any practice location, with optional daily/weekly recurrence. Each window can optionally be tagged with a list of note types via the UI; the selection is persisted (and surfaced back on edit) but is **not currently enforced by the slot filter** — see Known limitations.
- **Native scheduler enforcement** — A protocol subscribed to `APPOINTMENT__SLOTS__POST_SEARCH` filters Canvas's slot list down to `(Available windows) − (Busy blocks)`. Slots outside configured availability never appear in the native scheduler — no custom booking modal required.
- **Fail-closed** — Providers without configured availability show zero slots in the native scheduler until availability is set up. Configure before relying on the filter.

## Who it's for

Practices that want clinician-controlled availability with enforcement in Canvas's native scheduling workflow, without adopting a parallel booking UI.

## Configuration

- `SCHEDULABLE_STAFF_ROLES` (required): JSON array or comma-separated list of staff role internal codes (e.g. `["MD", "NP"]`) that should appear as bookable providers.
- Optional `BRAND_*` secrets re-skin the UI without forking — see `utils/theming.py`.

## Known limitations

- RRULE coverage is `DAILY` and `WEEKLY` only; `MONTHLY` / `YEARLY` patterns and `EXDATE` exceptions are not honored.
- The slot filter relies on a resolvable location in the `APPOINTMENT__SLOTS__POST_SEARCH` payload; when one is not present the filter passes slots through unchanged and emits a warning log.
- **Per-note-type restrictions are not enforced.** The UI lets you tag each window with a list of note types and the selection persists, but the filter does not consult a note type from the slot-search context. The full event payload shape is undocumented; the on-event log in `AvailabilitySlotFilter.compute` prints `selected_values` verbatim so we can identify (and wire in) the right key from real traffic.
- Drag-and-drop reschedules, API-created appointments, and other paths that bypass the slot picker are not validated.

## Test plan

- [ ] Install on a Canvas instance and confirm "Manage Availability" appears in the provider/hamburger menu
- [ ] Create Available + Busy windows for a provider, including recurring (daily/weekly) and one-off events
- [ ] Verify the native appointment book hides slots outside Available windows
- [ ] Verify the native appointment book hides slots that fall inside a Busy block
- [ ] Verify a provider with no Available windows shows zero slots in the native scheduler
- [ ] Run \`uv run pytest tests/\` and confirm all tests pass


amandacanvas:
Fix pushed (0a37de4).

**Bug:** `main.js` had 14 fetch URLs hardcoded with the sibling plugin's name (`scheduling_with_rooms`) instead of `provider_availability_manager` — leftover from when this plugin was scaffolded from `scheduling-with-rooms`. Since Canvas routes plugin APIs as `/plugin-io/api/<manifest.name>/...`, every save / patch / delete request landed on a non-existent route and the UI showed "Save failed: HTTP 404" on `decipherhealth-dev`.

**Fix:** Replaced the URL prefix on all 14 sites; the rest of the JS is unchanged. Bumped `plugin_version` 0.1.0 → 0.1.1 so the customer install is an unambiguous upgrade.

**Verification:** 138 tests pass, manifest validates.

amandacanvas:
@claude review once

amandacanvas:
Pushed c463711. Both review findings addressed.

### 1. `_staff_calendars` admin substring collision
Kept the DB query broad (`description__icontains=<uuid forms>` OR `title__istartswith=<name>: `) and moved the type scoping into Python via `parse_calendar_title` — so the keyword only compares against the parsed *type* slot, not anywhere it happens to appear in the *location* slot. Pattern matches what `get_availability_windows` already does for location matching.

Three regression tests added in `test_calendar_availability.py`:
- `test_staff_calendars_does_not_match_admin_substring_in_location_name` — locks in that `"Dr Smith: Clinic: Admin Office"` is **not** returned for `type_keyword="admin"`
- `test_staff_calendars_finds_clinic_calendar_at_admin_office_when_searching_clinic` — the inverse, so the fix doesn't accidentally exclude the legitimate match
- `test_staff_calendars_matches_type_case_insensitively` — title casing variants still match

The existing `test_staff_calendars_dedupes_by_id` was updated to set explicit titles on its mocks (the old version's MagicMock titles silently bypassed the parser).

### 2. `allowed_note_types` documented but not enforced
Chose to remove the promise rather than ship a defensive guess at the undocumented `APPOINTMENT__SLOTS__POST_SEARCH` payload — risk of "the filter looks like it works but doesn't" was too high without ground truth from real traffic.

- Dropped the "well-child only" sentence from `README.md` (Availability manager UI section) and replaced with a note that the selection persists but isn't enforced yet.
- Added an explicit Known Limitations entry calling out what's stored vs not enforced, why (undocumented event shape), and how to recover the missing key (the existing log line in `compute` prints `selected_values` verbatim).
- Updated the PR description to match: removed "per-note-type gating" from the Summary bullet, removed the corresponding test-plan checkbox, added the limitation to Known Limitations.

UI chips and the `allowed_note_types` storage are unchanged — the data still flows through and can be wired in once we see a real payload.

### Verification
141 tests pass (138 existing + 3 new regression tests). Manifest validates.

amandacanvas:
@claude review once

amandacanvas:
Pushed aa3b50f. Both findings addressed.

### 1. `event_occurs_on_date` tz mismatch
`Event.starts_at` is stored as tz-aware UTC by `canvas_sdk` + Django `USE_TZ`. The function was calling `.date()` directly — returning the UTC date — and comparing against `target_date`, which the slot filter derived from the slot's *local* date. Evening events in PST/MST/CST/EST silently fell off the wrong UTC day and got dropped by the fail-closed branch.

Added `calendar_tz` parameter (defaults to UTC for backward compat with the existing naive-datetime fixtures) and converted with `.astimezone(calendar_tz).date()` at all 4 comparison sites (non-recurring equality, start guard, DAILY interval math, WEEKLY interval math). Production callers in `get_availability_windows` and `get_blocking_calendar_events` now pass the calendar's tz explicitly — matching what `_event_window_on_date` already does at line 119.

### 2. `_resolve_location_name` UUIDField crash
`PracticeLocation.id` is a UUIDField; Django raises `ValidationError` during query construction for non-UUID input. The two `PracticeLocation.objects.filter(id=…)` calls in `_resolve_location_name` had no guard, so a non-UUID value in `selected_values` would crash `compute()` and Canvas would receive no effect — flipping the README's fail-closed guarantee into fail-OPEN.

Wrapped both lookup sites in `try/except (ValueError, ValidationError)` and fall through to the existing fallback. Catching both exception types defends against future Django version churn (older Django raised `ValueError`, newer raises `ValidationError`).

### Regression tests
- `test_event_occurs_on_date_pst_evening_matches_local_monday` — uses real `tzinfo=timezone.utc` on `event.starts_at` and PDT for `calendar_tz`. Asserts True for the local Monday and False for the UTC Tuesday — exactly the production shape that the existing tests missed.
- `test_event_occurs_on_date_pst_evening_recurring_first_occurrence_not_missed` — same shape, but recurring. Locks the start-guard path.
- `test_event_occurs_on_date_tokyo_morning_matches_local_tuesday` — positive-offset mirror.
- `test_event_occurs_on_date_daily_interval_uses_local_dates` — DAILY/INTERVAL arithmetic uses the local delta.
- `test_non_uuid_string_location_falls_through_without_crashing` — the string-shape branch.
- `test_non_uuid_dict_id_falls_through_without_crashing` — the dict-shape branch.
- `test_value_error_from_filter_also_caught` — the older Django exception variant.

148 tests pass (141 prior + 7 new). Manifest validates.

### Same bug class observed elsewhere (not in this PR's scope)
`api/events.py:_calendar_tz` (line 25) and `_calendar_tz_for_event` (line 36) make the same `Calendar.objects.filter(id=<string>)` and `EventModel.objects.filter(id=<string>)` calls on UUIDField columns with arbitrary user input (request body / path param). I left these out of this commit to keep the diff narrowly scoped to the review findings, but flagging so we don't get a third pass on the same pattern — happy to follow up in this PR or in a separate one.

amandacanvas:
@claude review once

amandacanvas:
Pushed 898eb05. Fixed all three new findings plus the carry-over I previously flagged as follow-up — you were right that "known issue" isn't a substitute for fixing it.

### 1. WEEKLY+INTERVAL math now Sunday-anchored to match the JS UI
The Python filter was counting rolling 7-day chunks from DTSTART; the JS UI's `weekStartOf` counts Sunday-anchored calendar weeks (RFC 5545 default). For DTSTART on a non-Sunday weekday with BYDAY including an earlier-in-the-week day, every match was inverted. Added a `_sunday_of(d)` helper that mirrors the JS anchor exactly (`d - timedelta(days=(d.weekday() + 1) % 7)`) and anchored both sides before the `// 7`. Regression test pins the reviewer's exact `FREQ=WEEKLY;INTERVAL=2;BYDAY=MO,TH` with DTSTART Tue 2026-05-05 case.

### 2. Type toggle no-op in edit mode
Smallest closing fix: disabled the Type toggle (Available / Block) when `modal.mode === 'edit'`. Tooltip explains "Delete and recreate to change type". This matches what `editThisAndFollowing` already gets right (explicit `ensureCalendar(provId, f.type, ...)` on the continuation series) and aligns with how most calendar apps handle type-of-event changes. The slot filter is fail-closed on calendar type, so leaving the silent no-op exposed could let a user think they marked a window Busy while slots stayed bookable — closed completely.

### 3 + 4. UUIDField crash class fixed everywhere
You explicitly called out the punt — fair. The three sites I'd left unfixed (one flagged in the latest review, two from my own follow-up note on aa3b50f) are now all wrapped:

| Site | Fix |
|---|---|
| `_resolve_staff` (calendar_availability.py) | `except (Staff.DoesNotExist, ValueError, ValidationError)` — returns None on malformed input, cached as None so subsequent lookups don't re-trigger |
| `_calendar_tz` (api/events.py) | `try/except (ValueError, ValidationError)` around `Calendar.objects.filter(id=...)`, falls back to UTC |
| `_calendar_tz_for_event` (api/events.py) | `try/except (ValueError, ValidationError)` around `EventModel.objects.filter(id=...)`, falls back to UTC |

Each got a `ValidationError` regression test AND a `ValueError` regression test (older Django raises one, newer Django raises the other — catching both defends against the SDK upgrading Django under us).

### Whole-plugin audit (this time really)
I re-grepped for both bug classes across the entire plugin:
- `.objects.{get,filter}(id=<str>)` on UUIDField with potentially-untrusted input: **all 5 sites are now wrapped** (`_resolve_location_name` ×2, `_resolve_staff`, `_calendar_tz`, `_calendar_tz_for_event`)
- `.date()` on `event.starts_at` or `ends_at` outside `_event_window_on_date`: all 4 inside `event_occurs_on_date` are now tz-converted

### Verification
156 tests pass (148 prior + 8 new regression cases). Manifest validates.

amandacanvas:
@claude review once

amandacanvas:
Pushed 3ff8d15. All five findings addressed. I read every site fully before touching anything this round, and every fix follows a sibling in the same file that already gets it right.

| # | Issue | Fix |
|---|---|---|
| 1 | WEEKLY without BYDAY matched every day in Python; JS + RFC 5545 require the start weekday only | Added `else` arm in `event_occurs_on_date` requiring `target_date.weekday() == local_start_date.weekday()` when BYDAY is empty. The old `test_..._weekly_no_byday_always_matches` only asserted Mon→True on a Mon DTSTART, so it passed whether the semantic was "every day" or "match start weekday" — renamed and rewritten to assert the right semantic across the whole week. |
| 2 | `seriesKeyOf` omitted start date → non-recurring one-offs with same provider/location/type/title/hh:mm collapsed into one logical series, silently affecting multiple dates on a single delete/edit | Added start date to the key when `rule.type === ''`. Recurring multi-provider rollup and this-only override grouping still work (their date components stay empty). |
| 3 | `onSave` missing `!isRecurring` guard that `doScopedDelete` already has → non-recurring multi-provider edits silently duplicate each provider's record on every save | Mirrored the guard: `scope === 'all' \|\| !isRecurring` routes to `editAll`. |
| 4 | `onDelete` missing `else if (scope === 'following')` branch (sibling `doScopedDelete` has it at line 970) → "this and following" delete via the modal removed only the clicked occurrence and silently created a continuation series; confirm message also lied | Added the branch mirroring `doScopedDelete` (`truncateBefore` the covering segment, no continuation). Fixed the confirm-message ternary so each scope reads correctly. |
| 5 | `splitOutOccurrence` continuation DTSTART = dayAfter shifted the WKST anchor for INTERVAL>1 rules, inverting every future occurrence's active/skipped parity | Walk forward from `dayAfter` using `expandOccurrences` against the ORIGINAL rule; first match becomes the continuation DTSTART. Matching set is closed under the original modulo-interval cadence so this preserves the anchor exactly. The function comment had already acknowledged this in a parenthetical ("or the next valid occurrence on/after dayAfter") — implementation now matches the comment. |

### Verification
- 156 tests pass; the one test whose name + assertion encoded the buggy "matches every day" semantic is fixed.
- Manifest validates.
- Re-grepped for similar bug patterns in this round's surface area:
  - **All scope-dispatch sites** (`onSave`, `onDelete`, `doScopedDelete`, `onTilePopupAction`) now use consistent `scope === 'all' || !isRecurring` ⊕ `scope === 'following'` ⊕ `scope === 'this'` branching.
  - **All confirm messages** match the corresponding code branch.
  - **All `splitOutOccurrence` callers** route through the anchor-walking continuation.
  - **`seriesKeyOf`** is the only consolidation key in the file (verified by grep).

amandacanvas:
@claude review once

amandacanvas:
Pushed 4850c2e. Owning the regression first — Issue 4 of this round (the anchor walk losing ~year of slots) was a bug I introduced in 3ff8d15. The walk passed midnight as both `rangeStart` and `rangeEnd` to `expandOccurrences`, which rejects occurrences at 09:00 because `09:00 <= 00:00` is false. I should have written a JS regression test for that walk; instead I'll lay the fix down clearly and the surrounding ones with it.

| # | Issue | Fix |
|---|---|---|
| 1 | (My regression) `splitOutOccurrence` anchor walk dropped every iteration → continuation DTSTART landed ~366 days out, fail-closed dropped a year of slots | Pass end-of-day rangeEnd (`23:59:59.999`) so 09:00 lands inside the per-day bracket. |
| 2 | `editAll` collapsed split recurring segments (A truncated, C continuation) onto records[0]'s date range, destroying C and creating overlapping duplicates on A's dates | New `eventBodyForPatchPreserveSegment(f, record)` helper preserves per-record start-date and per-record `recurrence_ends_at`. `editAll` PATCH path uses it. Title / hh:mm / recurrence rule / allowed_note_types still broadcast (those are what "Edit > All" means). |
| 3 | `buildRecurrencePayload` hardcoded `endsAt: null` for preset modes; only `custom` read `customEnds`/`customEndDate`. `inferRecurrenceMode` routed finite weekly series to `weekly_day` regardless of endDate → silent open-ending on Edit > All / Following | Compute `presetEndsAt` from the same form fields the custom branch reads; thread through all three preset branches. |
| 4 | Provider chip narrowing not enforced. Scope prompt narrowed `seriesRecords` to one provider, but chips rendered from full `state.providers` — clicking a previously-in-series chip created a duplicate event | `openEditModal` computes `allowedProviderIds` when `recordsOverride` is a strict subset; stores on `state.modal`. `makeChips` greys out chips outside that set with a tooltip + short-circuits the click handler. Mirrors the Type-toggle disabled-in-edit pattern. |

### Verification
- 156 tests pass (Python slot filter, downstream of every JS bug)
- Manifest validates
- Hand-traced each fix against the failure scenarios in the review:
  - PST 9-11 weekly Monday + "this only" override on a Monday → continuation now DTSTART = next Monday, not 366 days out
  - Split-segment Edit > All (records[0]=A, records[1]=C) → A keeps 5/4-5/17, C keeps 5/25-open, only title broadcasts
  - Finite weekly_day series with customEndDate=2026-12-31 + Edit > All → PATCH carries `recurrenceEndsAt: "2026-12-31T23:59"` (was null)
  - Scope prompt → "Edit for Dr A only" + click Dr B's chip → click ignored, tooltip explains

amandacanvas:
@claude review once

amandacanvas:
Pushed d66de4b. Three reviewer findings + one proactive fix in the same bug class.

### Reviewer findings
| # | Issue | Fix |
|---|---|---|
| 1 | `openEditModal` used `records[0]` instead of the record matching the clicked `masterId`. Click in C's range populated form from A → A's stale truncation cap flowed onto the wire as `recurrenceEndsAt` via 4850c2e's preset plumbing → "born dead" POST with DTSTART > UNTIL → provider's schedule silently vanished | One-line: `ref = records.find(r => r.id === masterId) \|\| records[0]` |
| 2 | `editAll` applied `eventBodyForPatchPreserveSegment` to every record unconditionally. The helper exists for split-segment case; over-application silently dropped `f.date` for non-recurring moves and `f.customEndDate` for single-segment recurring extends/shortens | Dispatch on `segments.length > 1` — preserve only for the actual split case, else form-derived |
| 3 | All three "this and following" sites used `.find(...recordCovers)` → leaked downstream segments from prior split overrides; also caused overlap-on-create in `editThisAndFollowing` | Factored `classifySegmentsForFollowing` that buckets per-provider records into `covering` (truncate) + `downstream` (DELETE — also handles the edge case where covering segment starts exactly on `occurrenceDate`, where truncating would yield a degenerate endDate). Wired into `doScopedDelete` 'following', `onDelete` 'following', and `editThisAndFollowing` |

### Proactive: `splitOutOccurrence` "born dead" guard
Same bug class as Issue 2. For a click inside an already-truncated segment (second "this only" override), my 4850c2e anchor walk would find no valid match before the in-memory `recurrence.endDate` cap, hit the 366 iteration limit, and POST a continuation a year out with DTSTART > UNTIL → zero occurrences. Two-part guard: (a) break the walk once `anchorDate > recEndDt`, (b) set `found = true` only when `expandOccurrences` actually returns a match; skip the POST entirely when nothing was found.

### Owning what I introduced
Two of the three reviewer findings (Issues #1 and #3) were caused by my 4850c2e commit:
- The over-applied `eventBodyForPatchPreserveSegment` (I fixed the split-segment collapse correctly but didn't think about what `!isRecurring → editAll` does to the same code path)
- The unfixed `records[0]` form-ref selection (I added `allowedProviderIds` to the modal in 4850c2e while ref selection was right next to it — should have caught the masterId mismatch in the same pass)

### How I'm changing the approach
You asked me to consider the whole bug class, not just the flagged sites. For this pass I traced through:
- Every `.find(recordCovers)` site in the file (6 total) — fixed the 3 'following' sites, verified the 3 'this only' sites are correct for their semantic
- Every consumer of `record.recurrence.endDate` in the split-series anchor logic — found and fixed the splitOutOccurrence "born dead" path proactively
- Every dispatch in `onSave` / `onDelete` / `onTilePopupAction` — verified the `editAll` change is consistent across both recurring and non-recurring routes

### Verification
- 156 tests pass (Python slot filter, downstream of every JS bug)
- Manifest validates
- Hand-traced 6 distinct failure scenarios from the review history end-to-end — all correct

amandacanvas:
@claude review once.

PR #287, opened 2026-05-19
Add vitals-dashboard plugin

allisonleitner:
## Summary

Adds the **vitals-dashboard** plugin: a Canvas full-chart tab for cardiology-grade vitals capture and review.

## What it does

Provides a per-patient \"Vitals\" tab where clinical staff can capture standard and orthostatic blood pressure with cuff location, current and dry weight, per-void urine output, O2 saturation, respiration rate, temperature, pain score, and edema notes. The dashboard renders 24h / 7d / 30d / 90d / all-time trend charts, supports CSV and Print export, and lets users carry forward values from the most recent finished session or mark committed measurements as Entered-in-Error.

Finishing a session emits a \`Vitals\` note with a read-only \`VitalsSummary\` custom command, then immediately syncs every captured measurement as a native FHIR \`Observation\` record via a \`POST /sync_observations\` SimpleAPI endpoint — so the vitals appear on the chart-summary sidebar and in CCDA exports.

Author: Allison Leitner.

## Test plan

- [ ] \`uv run pytest tests/\` passes (122 tests, ~94% branch coverage on installation)
- [ ] Install with \`canvas install vitals-dashboard\` on a clean instance
- [ ] Verify the Vitals tab appears under \`full_chart\` scope
- [ ] Enter a session (standard BP + orthostatic + weight + urine output), Save Session, reload — draft auto-restores
- [ ] Click Finish Session — new Vitals note appears with VitalsSummary command; toast confirms observation sync; FHIR \`Observation\` count increments
- [ ] Carry Forward Last Values from a prior finished session
- [ ] Edit a measurement (PATCH) — verify cross-patient dbid walks return 404
- [ ] Mark a measurement as Entered-in-Error — row remains visible but excluded from charts and CSV
- [ ] Print Report renders with practice header + chart snapshots; CSV export contains the cuff-location column

## Notes for reviewers

- The plugin stores its own \`VitalsSession\` and \`VitalsMeasurement\` rows under namespace \`vitals__dashboard\`; \`Observation\` records are written in parallel for native chart-summary / CCDA visibility.
- Screenshots TODO is intentionally left in the README — happy to add a recording link once a reviewer points at a preferred location for plugin media.


allisonleitner:
@claude review

amandacanvas:
@claude review once

amandacanvas:
@claude review once

allisonleitner:
Addressed both review findings in 18d9e3b:

**1. Stored XSS in `_render_summary_html`** — added `import html` and wrapped every user-controlled interpolation (`display_dt_str`, urine `value_text`, edema `value_text`) with `html.escape()`. For defense-in-depth I also escaped `_fmt_num` outputs and `CUFF_LABEL` lookups so future schema changes can't reopen the sink. Added a regression test (`test_user_text_is_html_escaped`) that pushes `` through all three sinks and asserts no raw tag survives.

**2. Blanket `try/except Exception` in `get_report_context`** — removed all four wrappers around `patient.telecom` / `patient.addresses` / `location.telecom` / `location.addresses`. These are normal prefetched ORM relations; per REVIEW.md, real failures should propagate to Sentry rather than be silently swallowed.

Tests: **123/123 passing** (was 122).

allisonleitner:
@claude review

allisonleitner:
Round-2 review addressed in 1dc9c79:

**1. Finished-session overwrite (vitals_api.py:481)** — `create_session` now returns **409 CONFLICT** when `update_session_id` targets a session with `note_id` populated, matching the contract already enforced by `/sync_observations`. Stale-tab autosaves can no longer silently rewrite measurements that back a signed note.

**2. Urine void timezone mismatch (vitals_api.py:413)** — frontend now sends `recorded_at_display` (raw HH:MM the clinician entered) per void; the server stashes it as a transient attribute on the in-memory row; the renderer prefers it over the UTC-shifted `recorded_at.strftime("%H:%M")`. No model migration. The session header and per-void rows now agree in every timezone.

**3. Silent staff fallback (vitals_api.py:666)** — dropped the `Staff.objects.first()` substitution. When the authenticated staff id doesn't resolve, `_build_finish_effects` now raises `_FinishError("Authenticated staff not found.")`, matching the existing patterns for missing `NoteType` and `PracticeLocation`. Misattributing a signed note to a real other clinician was strictly worse than failing fast.

**4. Frontend XSS in printReport() (vitals_dashboard.py:1111)** — wrapped `fmtValue(r)`, `entered_by`, and `provider_of_record` interpolations with `esc()`, matching the pattern already used throughout the print template and in `renderRowView`. The server-side escape from round 1 didn't cover this popup path; the same urine/edema payload could still fire in the about:blank popup (which inherits the opener's Canvas origin).

Tests: **125/125 passing**. Added regression tests for urine-time-display, urine-time-UTC-fallback, finished-session-409, and rewrote the staff-fallback test to assert the error instead of the substitution.

allisonleitner:
@claude review

allisonleitner:
Round-3 review addressed in d9ca26a:

**1. Cross-patient note_id linking (vitals_api.py:508)** — `create_session` no longer accepts a body-supplied `note_id` on the create path. The legitimate finish flow sets it server-side from the just-generated uuid4; the frontend never sent the field anyway. This closes the vector where an authenticated caller could plant a victim patient's note UUID and then have `/sync_observations` emit Observation effects with mismatched `(patient_id, note_id)` linkage.

**2. Soft-delete destruction in draft-update path (vitals_api.py:500)** — added `is_deleted=False` to the bulk-delete filter so Entered-in-Error rows survive a subsequent autosave. The previous unfiltered `.delete()` was silently wiping the audit row the confirm dialog had promised would be preserved.

**3. Missing finished-session guard on PATCH (vitals_api.py:987)** — `update_measurement` now resolves the parent `VitalsSession` and returns **409 CONFLICT** when `note_id` is populated, mirroring the round-2 guard on `create_session`. Without this, editing a measurement that backs a signed Vitals note silently diverges the dashboard/CSV/Print Report from the signed `VitalsSummaryCommand` HTML + emitted FHIR Observations.

**4. README accuracy on Entered-in-Error scope** — the README's bullet now states explicitly that retraction is dashboard-scoped (excludes from this plugin's charts, CSV, Print Report, and audit strikethrough) and does **not** propagate to the FHIR `Observation` already emitted to chart-summary / CCDA. Recommends starting a corrected session for chart-wide changes. The durable fix (persisting `Observation.id` per `VitalsMeasurement` and setting `entered_in_error_id` at delete time) is out of scope for this PR — no current Canvas SDK pattern exposes that field cleanly, and the sibling plugins (`vitals_visualizer`, `custom-observation-management`) only read on `entered_in_error_id__isnull` rather than write it.

Tests: **128/128 passing**. Added body-`note_id`-ignored, soft-delete-survives-update, and PATCH-finished-session-rejected regression tests; updated the two existing tests that asserted the old behavior.

Also ran the MSF OSS-readiness checklist locally — clean on security (no PHI / no secrets / no customer URLs / no internal-system references), git history is clean, LICENSE + `pyproject.toml` license = "MIT", README has all six required sections, manifest name + description satisfy length rules. Only outstanding OSS-checklist item is the screenshot file in `extensions/vitals-dashboard/` — flagged in the PR body and pending a reviewer's preferred location.

allisonleitner:
@claude review

PR #293, opened 2026-05-19
order-sets: load modal from API endpoint directly

amandacanvas:
## Summary

Simplifies how `OrderSetsApp` and `OrderSetsAdminApp` open their modals. Replaces the inline-HTML-shell pattern with a direct `LaunchModalEffect(url=...)` that points at the plugin's SimpleAPI endpoint.

**Before:** Canvas uploaded the on_open content HTML to S3, the browser loaded a pre-signed URL, that page's inline `<script>` fetched the real UI from `/plugin-io/api/order_sets/ui`, and `document.open/write/close` swapped the loading shell for the real UI.

**After:** Canvas's iframe loads `/plugin-io/api/order_sets/ui` (or `/admin-ui`) directly in a single request. The `/ui` and `/admin-ui` endpoints already return full `HTMLResponse` pages, so no other changes were needed in the API layer.

### Why

- Eliminates the S3 upload + pre-signed URL roundtrip (and the short-lived URL expiry that came with it).
- Drops the inline `<script>` / `document.write()` pattern.
- Removes the `_js_safe` HTML-tokenizer-aware escape helper — URL-encoded query parameters do the equivalent job, and `urllib.parse.quote` is a standard-library primitive rather than a hand-rolled escape.
- Roughly 60 lines of code removed across the two handlers.

### Tests

`uv run pytest tests/` — 146/146 passing.

`TestOrderSetsApp` and `TestOrderSetsAdminApp` were updated to assert the new `url` field on the resulting `LaunchModalEffect` instead of inspecting the inline HTML. The script-tag-breakout test was replaced with a percent-encoding test that verifies reserved URL characters in `patient_id` are escaped via `quote(..., safe='')` before being interpolated into the query string.

## Test plan

- [ ] `uv run pytest tests/` passes locally
- [ ] Install on a Canvas instance and confirm Order Sets opens in the right chart pane from a patient chart
- [ ] Confirm Order Sets Admin opens in the default modal from the global app drawer
- [ ] Confirm no console errors or S3-related network failures on click

### DIFF
--- a/extensions/order-sets/order_sets/applications/order_sets_admin_app.py
+++ b/extensions/order-sets/order_sets/applications/order_sets_admin_app.py
@@ -14,30 +14,8 @@ class OrderSetsAdminApp(Application):
     """
 
     def on_open(self) -> Effect:
-        html = (
-            '<html><head><style>'
-            'body { font-family: -apple-system, sans-serif; background: #f9fafb; margin: 0; padding: 0; }'
-            '.loader { display: flex; justify-content: center; align-items: center; min-height: 100vh; flex-direction: column; color: #6b7280; }'
-            '.spinner { width: 40px; height: 40px; border: 4px solid #e5e7eb; border-top-color: #4F46E5; border-radius: 50%; animation: spin 0.8s linear infinite; margin-bottom: 16px; }'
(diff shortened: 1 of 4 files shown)


amandacanvas:
@claude review once

PR #294, opened 2026-05-19
order-sets: fix 404 on plugin API by registering under protocols

amandacanvas:
## Summary

Fix: the `OrderSetsAPI` SimpleAPI handler was declared under `components.handlers` in the manifest. The plugin runner in canvas-plugins only loads handler classes from `components.protocols` and `components.applications` — the `components.handlers` array is silently ignored at routing time.

Result: even though install succeeded and the apps registered correctly, requests from the iframe to `/plugin-io/api/order_sets/ui` and `/plugin-io/api/order_sets/admin-ui` returned 404, and clicking the Order Sets app icons opened blank modals.

### Reference

`canvas-plugins/plugin_runner/plugin_runner.py:590`:

```python
handlers = manifest_json["components"].get("protocols", []) + manifest_json[
    "components"
].get("applications", [])
```

Other SimpleAPI plugins in this repo that register without a `PREFIX` (e.g. `scheduling-with-rooms`) already use `components.protocols` — this PR aligns `order-sets` with that pattern.

### Tests

`uv run pytest tests/` — 146/146 passing. No code changes; tests verify the manifest is internally consistent and don't depend on its top-level section.

## Test plan

- [ ] Reinstall the plugin on a Canvas instance
- [ ] Click the Order Sets app icon in a patient chart → modal opens with the order-set browser UI
- [ ] Click the Order Sets Admin app icon in the global drawer → modal opens with the admin UI
- [ ] Verify in DevTools Network: GET `/plugin-io/api/order_sets/ui?patient_id=…` returns 200 with HTML
- [ ] Verify in DevTools Network: GET `/plugin-io/api/order_sets/admin-ui` returns 200 with HTML

PR #290, opened 2026-05-19
Add order-sets plugin

amandacanvas:
## Summary

Adds an **Order Sets** plugin that lets clinical staff select predefined bundles of lab, imaging, and point-of-care (POC) in-office orders from the patient chart. Providers can apply an entire order set with one click, or preview and deselect specific items before confirming.

The plugin is workflow-agnostic and works for primary care, urgent care, weight management, behavioral health, and specialty practices that have repeatable order panels.

### What's included

- **Patient-scoped application** (`order_sets.applications.order_sets_app:OrderSetsApp`) — in-chart browse / order flow
- **Global-scope admin application** (`order_sets.applications.order_sets_admin_app:OrderSetsAdminApp`) — manage practice-wide and personal sets without a patient context
- **SimpleAPI handler** (`order_sets.api.endpoints:OrderSetsAPI`) — staff-authenticated REST endpoints for CRUD and order execution
- **Authorization model**:
  - Personal sets: editable only by their creator
  - Shared sets: editable only by staff with an Administrative `StaffRole`
  - Non-providers (MAs, nurses) must explicitly select an ordering provider before placing orders
- **Dynamic catalogs**: lab tests pulled from the instance's configured `LabPartner`s; POC CPT codes from `ChargeDescriptionMaster`

### Authorship

Amanda Peterson <amanda.peterson@canvasmedical.com>

## Test plan

- [ ] `uv run pytest tests/` passes (146 tests, all green locally)
- [ ] Install on a clean Canvas instance: `canvas install order_sets`
- [ ] Patient chart: open the **Order Sets** app from the patient app drawer; verify shared and personal sets render
- [ ] Quick-order a set against an open note; verify lab / imaging / perform commands are created
- [ ] Preview a set, deselect one item, confirm; verify only selected items are ordered
- [ ] Global drawer: open the **Order Sets Admin** app; create / edit / delete a shared set and a personal favorite
- [ ] Authorization: as a non-admin, confirm you cannot edit or delete a shared set
- [ ] Authorization: as a non-provider, confirm ordering requires explicit provider selection

PR #295, opened 2026-05-19
order-sets: align manifest with working MSF plugin pattern

amandacanvas:
## Summary

After PRs #290, #293, and #294, the plugin still wasn't launching modals on the instances it was tested on. Doing a side-by-side comparison against every Pattern C MSF plugin that uses `LaunchModalEffect(url=)` with a no-PREFIX SimpleAPI (`scheduling-with-rooms`, `intake-assignment-panel`, `provider_clinical_summary_companion`, `provider_note_vitals_companion`, `encounter_list`) surfaced several remaining ways this plugin's manifest and file layout diverged from their consistent shape. This PR aligns all of them.

## Changes

| # | Change | Why |
|---|---|---|
| 1 | Add `"url_permissions": []` at top level | Every working analog has this set to `[]`; we were the only one missing it entirely |
| 2 | Shorten `data_access.read` to bare class names (`"LabPartner"` instead of `"canvas_sdk.v1.data.lab.LabPartner"`) | Matches the convention every working plugin uses |
| 3 | `sdk_version`: `0.123.1` → `0.1.4` | Matches `scheduling-with-rooms`; previous value was higher than anything observed in working MSF plugins |
| 4 | Move `README.md` from outer plugin dir into the inner `order_sets/` package; manifest `readme` field changes from `"../README.md"` to `"./README.md"` | The `../` traversal pointed outside the uploaded archive when installing via URL |
| 5 | Update outer `pyproject.toml` `readme` reference to `"order_sets/README.md"` | Python tooling continues to find the file in its new location |

## Not changed

- `custom_data.namespace` — load-bearing; the plugin's `OrderSet` model is a `CustomModel` that uses this namespace
- Application class code, API endpoint code, on_open URL construction — all unchanged

## Tests

`uv run pytest tests/` — 146/146 passing. No code changes; this is purely manifest and layout hygiene.

## Test plan

After merging, the clean reinstall procedure is:

1. In Canvas admin, uninstall **any** plugin matching `"Order Sets"`, `"order_sets"`, or `"order-sets"` — across iterations there may be multiple registered
2. Hard-refresh the Canvas page (Cmd+Shift+R) to flush cached plugin metadata
3. Install fresh from `https://github.com/Medical-Software-Foundation/canvas/tree/main/extensions/order-sets/order_sets`
4. Verify in the plugin admin list that exactly **one** `order_sets` entry exists and reports no errors
5. Click an Order Sets app icon and confirm the modal opens with content


amandacanvas:
@claude review once

PR #296, opened 2026-05-20
Remove order-sets plugin

amandacanvas:
## Summary

Removes the `order-sets` plugin from the public extensions directory. The plugin should not be available for customer install in its current state.

## Test plan

- [ ] CI passes
- [ ] After merge, `extensions/order-sets/` is no longer present on `main`
- [ ] After merge, the plugin is no longer installable via the Canvas Upload-from-URL flow pointing at this repo


amandacanvas:
@claude review once

PR #300, opened 2026-05-20
Add compound-medication-loader plugin

allisonleitner:
## What this plugin does

Bulk-loads Compound Medications into a Canvas instance via either:

- **App-drawer page** — staff drag-drop a CSV, review the parsed rows (with errors and duplicates flagged *before* submission), and click Load.
- **Authenticated POST endpoint** — programmatic loading from a script via a per-instance bearer token, for one-off migrations or syncing from an external source.

Both paths share the same row validation, dedup logic, and emit `CREATE_COMPOUND_MEDICATION` effects.

## Who it's for

Clinical and admin staff at practices maintaining a compound-medication formulary — functional-medicine clinics, pain-management practices, ketamine clinics, dermatology, etc. Adding compounds one-at-a-time through Prescribe is tedious; this lets staff load the whole formulary from a CSV and re-run safely without duplicates.

## Author

Allison Leitner — `allison.leitner@canvasmedical.com`

## Test plan

- [ ] `uv run pytest tests/` passes locally (33 tests, 100% line + branch coverage)
- [ ] Install on a Canvas instance and confirm the "Compound Medication Loader" entry appears in the app drawer
- [ ] Drop a small CSV with new compounds — verify rows show "Ready" status, click Load, confirm rows persist
- [ ] Re-upload the same CSV — verify rows are flagged "Already in Canvas (active)" with yellow row tint
- [ ] Drop a CSV with an invalid `potency_unit_code` — verify row shows "Error" with the validation message
- [ ] Toggle "Skip rows whose formulation already exists" off — verify the Load count updates accordingly
- [ ] For programmatic flow: set the `BULK_LOAD_API_KEY` plugin secret, POST a one-row payload with `Authorization: Bearer <key>` — verify summary response shape and that the row appears in Canvas's Prescribe search

## Notes for reviewers

- Plugin uses absolute imports throughout (`from compound_medication_loader.api.loader import ...`)
- No try/except blocks except where validating user-provided JSON or row data — exceptions otherwise propagate to Canvas logs
- The `CREATE_COMPOUND_MEDICATION` effect is constructed directly with a flat JSON payload rather than via `CompoundMedication(...).create()` because the server-side interpreter validates against a flat schema (see comment in `api/loader.py:_process_row`)
- README has one outstanding `<!-- TODO -->` placeholder for screenshots


allisonleitner:
@claude review

allisonleitner:
Thanks for the careful review — addressed all three findings in [`e53987d`](https://github.com/Medical-Software-Foundation/canvas/pull/300/commits/e53987d):

- **🔴 controlled_substance default-to-N removed** at both the Python validator (`api/loader.py:_validate_row` + `_process_row`) and the JS validator (`templates/loader.html`). A missing column or blank cell now surfaces as a per-row error rather than silently classifying as Not Scheduled. Two new tests cover this: `test_missing_controlled_substance_row_errors` and `test_blank_controlled_substance_row_errors`.
- **🔴 Blanket `try/except Exception` around `Effect()` dropped.** Effect is internal pydantic over validated data — failures should propagate to Sentry/Canvas logs. Removed the test that exercised the deleted swallow path.
- **🟡 Bearer compare** switched to `hmac.compare_digest` to match the repo convention.

Regression-verified against allison-training: happy-path row still creates, row missing `controlled_substance` now rejects with the new validation error, bad bearer still 401. 35/35 tests pass, 100% line + branch coverage maintained.

Still outstanding: screenshot in the README.

allisonleitner:
@claude review

PR #301, opened 2026-05-20
Add order_sets extension

amandacanvas:
## What this plugin does

`order_sets` lets providers apply predefined bundles of lab, imaging, and point-of-care (POC) orders to a patient chart with a single click — instead of adding tests one at a time. Sets can be practice-wide (shared, admin-managed) or personal to each provider.

## Who it's for

Primary care, urgent care, and specialty clinics whose providers repeatedly order the same group of tests (e.g. diabetes workup, sports physical lab panel, standard chest workup). Admins curate shared sets; individual providers maintain their own favorites.

## Components

- `OrderSetsApp` — patient-specific Application launched from the 9-box app drawer
- `OrderSetsAPI` — SimpleAPI handler with endpoints for CRUD on order sets and order execution. Persists order-set definitions through the plugin cache.
- Vanilla-JS frontend (browser + admin views) with templates and static assets

Execution wires up `LabOrderCommand`, `ImagingOrderCommand`, or `PerformCommand` depending on the order type, attached to the patient's open note. Configurable ICD-10 diagnosis codes and comments flow through to every order in the set.

## Author

Amanda Peterson <amanda.peterson@canvasmedical.com>

## Test plan

- [ ] \`uv run pytest\` — 57 tests pass, 96% coverage
- [ ] \`uv run mypy --config-file=mypy.ini .\` (locally; mypy config not committed) — clean
- [ ] Install on a clean Canvas instance via \`canvas install order_sets\`
- [ ] Open a patient chart, click the **Order Sets** icon, confirm the right-pane launcher loads
- [ ] Create a shared order set with at least one lab item, apply it to an open note, verify a \`LabOrderCommand\` is placed
- [ ] Create an imaging order set, apply it, verify one \`ImagingOrderCommand\` per item
- [ ] Create a POC order set, apply it, verify \`PerformCommand\` is placed for each item
- [ ] Verify a personal (non-shared) set is only visible to its creator
- [ ] Verify execution blocks gracefully when no open note exists on the patient
- [ ] Verify provider fallback: explicit provider selection works when the open note has no provider attached


amandacanvas:
@claude review once

amandacanvas:
Addressed all four 🔴 review items in 233cf4f. Quick map from review comment → fix:

| # | Review finding | Fix |
|---|---|---|
| 1 | N+1 via `role.staff.id` in `list_providers` + `_resolve_provider` | `list_providers` now reads `role.staff_id` (the FK column on the row). `_resolve_provider` is rewritten as a single `.exists()` filtered on `(role_type, staff_id)` — zero row iteration. |
| 2 | `create_set` persists `created_by=""` when staff is `None` | Returns **401** before any write when `_current_staff()` is `None`. `created_by` is now always the authenticated staff id. |
| 3 | Blanket `try/except Exception` across 7 handlers swallows Sentry errors | Removed the body-wrapping `try/except` in `list_sets`, `delete_set`, `list_providers`, `get_note_provider`. Narrowed `create_set` / `update_set` to catch **only** `ValueError` from `request.json()` (→ 400). Inlined `_is_active_staff` (its try/except was dead). |
| 4 | `update_set` / `delete_set` allow any staff to overwrite shared sets | New `_can_modify` check: caller must be `created_by` **or** be in the new `ADMIN_STAFF_IDS` plugin secret. Fails closed when the secret is unset (with `log.warning`) — same pattern as `dea-prescriber-filter`. Defense-in-depth: an empty `created_by` never matches an empty caller id. |

`ADMIN_STAFF_IDS` is now declared in `CANVAS_MANIFEST.json` and documented in the README's Configuration section, including the authorization model (create / read / update-delete).

Test changes:
- 57 → 68 tests (added 11 covering the auth gates + an N+1 regression test that booby-traps `role.staff` so a future regression fails loud)
- The four swallow-error tests became `pytest.raises(...)` so the rule "errors must reach Sentry" is now exercised
- 99% coverage (up from 96%), mypy clean

@claude review once

amandacanvas:
@claude review once

amandacanvas:
Pushed 689937a. Two follow-ups addressed:

### 1. Storage: cache → `OrderSet` CustomModel

You're right that the 14-day TTL cache was the wrong tool here. Order-set definitions are reference data — clinics curate them over time and expect them to persist. Switched to a dedicated `OrderSet` CustomModel:

- `order_sets/models/order_set.py` — schema with `set_id` (UUID, indexed) plus a composite index on `(is_shared, created_by)` for the `list_sets` visibility query
- `CANVAS_MANIFEST.json` — declared `custom_data.namespace = "custom__order_sets"` with `read_write` access (mirrors `patient-tags`)
- `endpoints.py` — `list_sets` now uses `OrderSet.objects.filter(Q(is_shared=True) | Q(created_by=staff_id)).order_by("name")` as a single indexed query instead of loading the whole list and filtering in Python. `create_set` uses `OrderSet.objects.create(...)`; `update_set` / `delete_set` look up by `set_id` and call `.save()` / `.delete()` on the instance after the existing auth check; `execute_set` / `execute_custom` query the model too. Dropped the cache helpers and the `json` import.

> 📝 **Heads-up on mydayclinic-dev:** existing 0.0.25 deployments have order-set data in the cache, not in the table. If/when they upgrade, the cache will be ignored and the table will start empty. Coordinate the upgrade with that customer before pushing.

### 2. `_resolve_provider` missing `Staff.active` check

Bot review caught this in the second pass. Added `staff__active=True` to the existence filter — without it, an old open note whose original provider has since been deactivated would still resolve, and the order would originate under the deactivated provider. Now consistent with `list_providers` (`filter(active=True)`) and `get_note_provider` (`provider.active`). The test asserts all three kwargs (`role_type`, `staff_id`, `staff__active`) are in the filter call.

### Other notes

- Restored `mypy.ini` at the extension level — `note-production-dashboard` and `patient-tags` both have one in the public repo, so the publishing skill's blanket "remove dev tooling" was too aggressive. Added a per-module `var-annotated` override for the models package (matches the `patient-tags` pattern).
- 67 tests passing, 99% coverage, mypy clean, manifest valid.

@claude review once

amandacanvas:
@claude review once

amandacanvas:
Pushed 6e97a93. All four issues addressed and — equally important — I changed how I'm working to stop generating new findings.

**Root cause of the whack-a-mole**: each prior fix was verified against a mock or a docstring, not against actual SDK semantics. This round I read `canvas_sdk.v1.data.staff.py` first, confirmed `Staff.id` is a 32-char hex `CharField` (via `uuid.uuid4().hex`) and `StaffRole.staff` is a `ForeignKey(Staff)` targeting `Staff.dbid`, and used `dea_prescriber_filter/engine/lookups.py` as the canonical reference for the provider query.

### Fixes
1. **Provider query**: `list_providers` and `_resolve_provider` now query from the Staff side via the `roles` reverse relation (`Staff.objects.filter(active=True, roles__role_type="PROVIDER")...`). The previous `role.staff_id` comparison was an int-vs-hex type mismatch that broke `/providers` and Quick Order in production. Matches `dea_prescriber_filter.engine.lookups`.

2. **`execute_set` / `execute_custom` authorization**: Added `_can_read(order_set, staff)` mirroring the `list_sets` predicate. Knowing a UUID no longer lets one user execute another's personal set against their patient. 403 on cross-user personal-set execution; shared sets remain open to any authenticated caller.

3. **`ADMIN_STAFF_IDS` canonicalization**: New `_canonical_id()` normalizes both sides through `uuid.UUID(...).hex` before comparison. Dashed and undashed UUIDs in the secret both match an undashed `Staff.id`. Non-UUID values pass through verbatim (legacy internal codes still work). README updated to note both forms are accepted.

4. **`confirmOrder()` TypeError**: Captured `currentPreviewSet.id` into a local `setId` before calling `closePreview()`. Custom Order via Preview now actually places the order.

### Tests
- **Switched all staff-id fixtures from arbitrary strings to `uuid.uuid4().hex`.** This is what masked the previous round's bugs — `"admin-1"` was string-equal on both sides, so the canonicalization gap and the dbid/hex mismatch slept through the tests.
- Added regression tests for: the Staff-side filter shape (asserts `active=True, roles__role_type="PROVIDER"`), `_resolve_provider` filter shape (asserts `id=, active=True, roles__role_type=`), execute 401/403/shared-allowed/creator-allowed, dashed-vs-undashed admin UUID acceptance, `_canonical_id` non-UUID fallthrough.
- 75 tests, 99% coverage, mypy clean, manifest valid.

@claude review once

amandacanvas:
@claude review once

PR #277, opened 2026-05-16
Add exam_chart_app plugin

ngreer99:
## Summary

Adds an "Exam" tab to the note body that walks providers through Reason for Visit, History of Present Illness, Review of Systems, Physical Exam, and Assessment & Plan (with diagnoses and orders for labs, imaging, prescriptions, referrals, goals, plan items, and follow-ups) in one guided form. Clicking "Finalize Exam" emits the corresponding Canvas commands onto the note.

## Problem it solves

Encounter charting today is a patchwork of free-text dictation and fragmented commands, with no consistent way to drive HPI content from the visit reason. This plugin gives the provider a single guided form whose output is native Canvas commands, with HPI prefill keyed off the Reason for Visit.

## Who it's for

- Primary user: MD / provider completing the encounter
- Secondary user (read-only context): MA who completes intake before handoff

## How it's structured

- `ExamChartingApp` — `NoteApplication` registered under `components.protocols` (note-tab launcher; kept out of the chart's 9-dot drawer)
- `ExamChartingAPI`, `ExamSearchAPI` — `SimpleAPI` handlers, both using `StaffSessionAuthMixin`
- `ExamSectionAdditionalFieldsHandler` — bridges plugin-stashed narratives back into the chart's metadata-form UI for ROS / PE commands
- `ExamNoteLifecycle` — cleans up draft state on note state changes
- Per-note draft state and the post-finalize "ever finalized" marker stored in `AttributeHub` under namespace `canvas__exam_chart_app`
- Tests: 175 cases, mypy strict on production code

## Configuration

All optional, with fail-soft defaults:

| Secret | Purpose |
|---|---|
| `exam-note-types` | Comma-separated note-type keywords for tab visibility |
| `ros-questionnaire-code` | Pin a specific Review of Systems questionnaire by `Questionnaire.code` |
| `pe-questionnaire-code` | Pin a specific Physical Exam questionnaire by `Questionnaire.code` |
| `exam-imaging-codes` | Override the bundled CPT imaging picklist |
| `icd10-search-url` | Override the ICD-10-CM lookup endpoint (server-side validated: HTTPS only, blocks loopback / RFC-1918 / AWS IMDS) |
| `namespace_read_write_access_key` | Runtime-managed; auto-issued on first install |

## Test plan

- [ ] `uv run pytest` passes (175 tests)
- [ ] `uv run mypy --config-file=mypy.ini .` clean
- [ ] Install on a Canvas instance: `canvas install exam-chart-app`
[...]

PR #303, opened 2026-05-20
Fix exam_chart_app sandbox ImportError on install

ngreer99:
## Summary

Plugin install on a real Canvas instance currently fails with:

```
ImportError: 'requests.exceptions' is not an allowed import.
```

The Canvas plugin runtime sandbox blocks `from requests.exceptions import RequestException` at module load — even though `canvas_sdk` depends on `requests` internally, the sandbox's import allowlist is per-name within `requests.*` and `requests.exceptions` isn't on the list.

This was introduced during PR #277 review when `search_medications` was narrowed from `except Exception` to `except (RequestException, JSONDecodeError, ValueError)`. Local pytest passes because the dev venv allows the import; the failure only surfaces at install on a real Canvas instance.

## Fix

Switch the narrow catch in `exam_chart_app/api/exam_search_api.py:search_medications` from `(RequestException, JSONDecodeError, ValueError)` to `(OSError, ValueError)`:

- `RequestException` inherits from `IOError`, which Python 3 aliases to `OSError`. `except OSError:` catches every real-world network failure the SDK can raise from this call.
- `JSONDecodeError` inherits from `ValueError`, so the existing `ValueError` branch covers malformed-JSON bodies.
- Programming bugs (`AttributeError`, `KeyError`, `TypeError`) still propagate to Sentry — the original goal of the narrow-catch refactor is preserved.

The two now-unused imports (`from requests.exceptions import RequestException` and `from json import JSONDecodeError`) are dropped.

The existing `test_medications_returns_empty_on_ontologies_network_error` test is updated to fire `OSError` directly — matches the production catch and removes the `requests`-package dependency from the test source. `RequestException` is still caught at runtime (because it inherits from `OSError`), so the test would also pass with a real `requests.exceptions.RequestException`, but `OSError` is the cleaner contract.

## Test plan

- [x] `uv run pytest` — 187 pass
- [x] `uv run mypy --config-file=mypy.ini .` — clean (28 files)
- [ ] Install on a real Canvas instance — confirms the ImportError no longer fires

### DIFF
--- a/extensions/exam_chart_app/exam_chart_app/CANVAS_MANIFEST.json
+++ b/extensions/exam_chart_app/exam_chart_app/CANVAS_MANIFEST.json
@@ -17,21 +17,13 @@
             },
             {
                 "class": "exam_chart_app.api.exam_api:ExamChartingAPI",
-                "description": "Form-state load/save + finalize handler for the Exam tab",
-                "data_access": {
-                    "event": "",
-                    "read": ["AttributeHub", "Condition", "Questionnaire", "Staff"],
-                    "write": ["AttributeHub"]
-                }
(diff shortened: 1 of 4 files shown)


ngreer99:
@claude review once

ngreer99:
@claude review once

PR #307, opened 2026-05-21
Fix exam_chart_app django.db sandbox ImportError (follow-up to #303)

ngreer99:
## Summary

Follow-up to PR #303. That PR was merged before this commit (`e14c23a`) was pushed, so the django.db sandbox-import fix never made it into main. Plugin still fails to install on a fresh Canvas instance with:

```
plugin-runner ERROR Error importing module 'exam_chart_app:exam_chart_app.api.exam_api:ExamChartingAPI'
  File "/plugin-runner/custom-plugins/exam_chart_app/api/exam_api.py", line 26, in <module>
    from django.db import DatabaseError, OperationalError
ImportError: 'django.db' is not an allowed import.

Plugin "exam_chart_app" version 0.1.0 loaded with errors (4/5 handlers loaded successfully)
```

Every `@api.get("/exam/...")` route returns an empty-body 404 because the module never imports → routes never register → the `ExamChartingAPI` handler is the 1/5 that fails to load. Verified on corgi-sandbox 2026-05-21.

## The trap

The Canvas plugin sandbox blocks `from django.db import DatabaseError, OperationalError` even though `from django.db.models import Q` works fine (the SDK uses it). The allowlist is per-name within `django.db.*` and the exception classes at the `django.db` level are not on it.

This is the same class of issue as the `requests.exceptions` trap fixed in PR #303 (`b082d33`), but `django.db.*` exception classes don't inherit from `OSError`, so the `OSError` substitution that worked for requests doesn't apply.

Local pytest passes because the dev venv allows the import; the failure only surfaces at first-install on a real Canvas instance.

## Fix

Class-name string filtering against a small set of stable Django exception names:

```python
_DB_EXCEPTION_NAMES = frozenset({
    "DatabaseError", "OperationalError", "IntegrityError",
    "InterfaceError", "DataError", "NotSupportedError",
})

try:
    set_narrative(cmd_uuid, narrative)
except Exception as exc:  # noqa: BLE001 — sandbox blocks the natural narrow catch
    if exc.__class__.__name__ not in _DB_EXCEPTION_NAMES:
        raise  # programming bug → propagates to 500 + Sentry
    log.exception(...)  # pages on-call for swallowed DB transients
```

Django's exception class names are stable public API. The narrow-catch intent (programming bugs propagate, DB transients get swallowed) is preserved via the name-match; `log.exception` inside the matched branch keeps Sentry visibility.

Two sites updated, both post-gate best-effort swallows in `finalize()`:
- `set_narrative(cmd_uuid, narrative)` flush
- `mark_finalized(note_uuid)` + `mark_ever_finalized(note_uuid)`

## Test plan

- [x] `uv run pytest` — 187 pass (same as before; existing tests fire `OperationalError` and `AttributeError`, both still hit the right branches via the name filter)
- [x] `uv run mypy --config-file=mypy.ini .` — clean (28 files)
- [x] Reinstall on corgi-sandbox — `Successfully loaded plugin "exam_chart_app", version 0.1.0 (5/5 handlers loaded)`
- [ ] Reviewer can install on any Canvas instance and confirm the Exam tab loads its CSS/JS without 404s

### DIFF
--- a/extensions/exam_chart_app/exam_chart_app/api/emitters.py
+++ b/extensions/exam_chart_app/exam_chart_app/api/emitters.py
@@ -263,15 +263,20 @@ def _emit_diagnosis_block(
     assessment: dict[str, Any] = raw_assessment if isinstance(raw_assessment, dict) else {}
     status_raw = str(assessment.get("status") or "").strip().lower()
     assess_narrative = str(assessment.get("narrative") or "").strip()
-    # Existing-condition entries always emit Assess (with condition_id),
-    # even with empty status/narrative — the chart needs the Assess row
-    # to surface the existing condition as touched on this visit. New
-    # diagnoses skip Assess when both fields are empty (provider may
-    # not have intended one).
-    if existing_condition_id or status_raw or assess_narrative:
+    # Emit Assess only for existing conditions. AssessCommand requires
+    # a ``condition_id`` (UUID of an existing Condition row); the SDK
+    # has no ``icd10_code`` alternative to link to a Condition that's
+    # about to be created in the same note by DiagnoseCommand. Emitting
+    # Assess for a new dx produces an orphaned "Assess Condition:" row
+    # with an empty condition slot on the chart (observed 2026-06-12).
+    # For new diagnoses, the Status + Narrative form fields are
+    # therefore dropped — the form hides those fields on new-dx cards
+    # (see 50_diagnoses.js renderDxList) so providers don't enter data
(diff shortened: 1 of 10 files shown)


ngreer99:
@claude review once

ngreer99:
@claude review once

ngreer99:
cerberus review

ngreer99:
cerberus review

ngreer99:
cerberus review

ngreer99:
cerberus review

PR #315, opened 2026-05-26
dea_prescriber_filter: make pharmacy/prescriber state validation configurable

alyssa-yandura:
## Summary
- Adds a `STATE_VALIDATION_ENFORCE` plugin secret that controls whether the plugin blocks prescriptions when the selected prescriber's license state doesn't match the pharmacy's state.
- When set to `false` / `0` / `no` (case-insensitive), the blocking validation is skipped. Dropdown state annotations remain so reviewers can still spot mismatches visually.
- Defaults to enforce, so any 0.3.0 installation that upgrades without changing configuration sees no behavior change. Bumps `plugin_version` to `0.4.0`.

## Test plan
- [x] Installed against a Canvas dev instance via `canvas install`; upgrade-in-place from 0.3.0 succeeds
- [x] Default behavior (secret unset): prescribing with a prescriber licensed in one state and a pharmacy in another still produces the existing `"Prescriber state (X) does not match pharmacy state (Y)."` blocking error
- [x] Don't-enforce behavior (secret set to `false`): same flow no longer produces the state-mismatch error; dropdown still annotates each prescriber with their license state
- [x] Auth regression check: non-prescriber staff still sees `"Not authorized to prescribe for this provider."` and sign actions are hidden, regardless of the new secret's value

### DIFF
--- a/extensions/dea_prescriber_filter/CANVAS_MANIFEST.json
+++ b/extensions/dea_prescriber_filter/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.94.0",
-    "plugin_version": "0.3.0",
+    "plugin_version": "0.4.0",
     "name": "dea_prescriber_filter",
     "description": "Filters prescribe actions based on NPI matching between user and selected prescriber",
     "license": "MIT",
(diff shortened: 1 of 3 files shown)


alyssa-yandura:
@claude review once

PR #327, opened 2026-06-03
Add external-calendar-busy-blocks plugin

amandacanvas:
## Summary

A new Canvas plugin that lets each provider subscribe Canvas to their personal calendar (Google, Outlook, Apple iCloud) via the calendar's secret iCal URL. Every 15 minutes a `CronTask` fetches each connected feed, parses busy events, expands recurring events, and writes them as "Busy" Admin blocks on the provider's Canvas schedule. Source titles never leave the personal calendar — Canvas users see only "Busy."

- Self-service config via a global-scope `Application` (paste-the-URL form / Disconnect button) with session-authoritative `staff_id` (read from the logged-in user header, never the request body)
- Hand-rolled ICS parser + RRULE expander supporting `FREQ=DAILY|WEEKLY|MONTHLY|YEARLY` with `INTERVAL`, `BYDAY` (incl. positional `1MO`/`-1FR`), `BYMONTHDAY`, `BYMONTH`, `UNTIL`, `COUNT`, `EXDATE`, `RECURRENCE-ID`, capped at 1000 instances per VEVENT. Unsupported features (`BYSETPOS`, `BYWEEKNO`, `WKST`, sub-daily `BY*`) are rejected with `RRuleUnsupported` and the VEVENT is dropped with a warning rather than producing wrong output. The plugin runner sandbox does not allowlist `icalendar` or `dateutil.rrule`, so the parser is implemented in-house using only allowlisted modules (`datetime`, `zoneinfo`, `re`, `dateutil.relativedelta`).
- 15-minute cron with `If-None-Match` / `If-Modified-Since` so unchanged feeds short-circuit at the HTTP layer. Safety guard skips deletions when a feed returns 0 events but prior imports exist (protects against transient empty responses from Google).
- Two `CustomModel`s (`StaffCalendarFeed`, `ImportedEvent`) hold the URL, sync state, and the ICS-UID → Canvas-Event-ID mapping used for diff reconciliation. Plugin pre-allocates UUIDs client-side for `Event.create()` so subsequent updates/deletes can reference the same id.
- URL redaction in logs (path-segment + named query-param masking); HTTPS-only validation on save.

Spec: `docs/superpowers/specs/2026-06-01-external-calendar-busy-blocks-design.md`
Plan: `docs/superpowers/plans/2026-06-01-external-calendar-busy-blocks.md`

## Test plan

- [ ] `cd extensions/external-calendar-busy-blocks && uv run pytest tests/ -v` — 78/78 pass locally
- [ ] Install plugin against a dev Canvas instance
- [ ] As a provider: open "Calendar Busy Blocks" from the global menu, paste a real Google iCal URL, click Connect; page renders "Connected"
- [ ] Wait for cron (or trigger manually); confirm Admin "Busy" blocks appear on the provider's schedule for upcoming busy times
- [ ] Modify an event in the source calendar; confirm Canvas event updates on next sync
- [ ] Delete an event from the source calendar; confirm Canvas event is removed on next sync
- [ ] Click Disconnect; confirm all Busy blocks are removed
- [ ] Smoke-check log output — no full ICS URLs visible (tokens redacted to `***`)

## Known follow-ups

Coverage is at 87% overall. Parser modules (`ics/parser.py` 93%, `ics/rrule.py` 79%, `ics/datetimes.py` 84%, `sync/cron.py` 88%) have uncovered defensive branches worth backfilling tests for. Routes, Application, and UI are at target.


amandacanvas:
@claude review once

amandacanvas:
@claude review once

amandacanvas:
@claude review once

amandacanvas:
@claude review once

amandacanvas:
@claude review once

amandacanvas:
Pushed `429799f` addressing the latest review round. All four findings were verified against the current head and fixed at the source, plus a structural backstop so this class of issue can't recur.

**Defensive-parsing fixes** (all route malformed input into the existing per-VEVENT drop-and-continue contract — no behavior change for valid feeds):

- **Colonless property line** (`extract_vevents`): the raise happened *before* the per-VEVENT guard, disabling the whole feed. Now wraps `parse_property_line`, poisons just that VEVENT, and continues.
- **RECURRENCE-ID override** (`_build_override_event`): a bad-TZID override DTSTART/DTEND no longer kills the entire series — it falls back to the base occurrence; a missing DTEND inherits the base duration instead of silently dropping the instance. Genuine per-instance cancellation (`_should_skip`) is preserved, and the two existing override tests still pass.
- **`INTERVAL=0` / `COUNT<1`** (`parse_rrule`): previously expanded into 1000 duplicate `(uid, recurrence_id)` rows → `IntegrityError` mid-tick. Now validated → `RRuleUnsupported` (drops just the VEVENT).
- **Non-integer `SEQUENCE`**: `int("abc")` raised `ValueError`, which escaped the `IcsParseError` guards. Now converted to `IcsParseError`.

**Structural backstop** (`SyncCron.execute`): each `_sync_feed` call is now wrapped in a per-feed `try/except` that logs the traceback (Sentry-visible — not silent) and continues. A single provider's feed can no longer abort the whole tick or skip the other feeds, regardless of which parse path raises next. This also double-covers the duplicate-key case above.

Regression fixtures/tests added for each (colonless line, `INTERVAL=0`, bad `SEQUENCE`, bad-TZID override, DTEND-less override, per-feed isolation). **117 tests pass (was 109).** Manifest still validates.

One note for context: the four prior comments in the 14:32 and 16:54 batches were re-posted against the older `c8190fe` commit and were already resolved in `3b44fcd`/`57e8858` (staff_id canonicalization, in-progress recurrences, `custom_data` namespace, RRULE yield-vs-iteration cap, and source-timezone BYDAY evaluation). The five items above are the genuinely new ones from the `57e8858`/`3b44fcd` review.

amandacanvas:
@claude review once

amandacanvas:
Pushed `a6a6fac` addressing the three findings from the `429799f` round. All verified and fixed, with two deliberate divergences from the literal suggestions — flagging them here so they don't read as ignored, since applying them verbatim would each introduce a new issue.

**F — MONTHLY/YEARLY day overflow (`_candidate_days_in_month`)**: fixed exactly as suggested. Dropped the `else last` clamp so an out-of-range day (e.g. a day-31 monthly rule in February, or Feb 29 yearly in a non-leap year) skips the month per RFC 5545 §3.3.10 instead of emitting a wrong-day block. `_expand_yearly`'s existing `or [dtstart.day]` + `ValueError` guard then correctly skips Feb 29 in non-leap years.

**G — double-`@` SSRF bypass (`_is_allowed_host`)**: fixed, but **not** via `urllib.parse.urlsplit`. The plugin sandbox's import allowlist only exposes `urlencode`/`quote`/`unquote` from `urllib.parse` (`urlsplit`/`urlparse` are blocked — that's why this code used a regex originally), so `urlsplit(url).hostname` would raise `ImportError` at plugin load. Instead, `_extract_host` now parses the authority and splits userinfo on the **last** `@`, then strips the port (RFC 3986) — so the allowlist reads the same host `requests`/`urllib3` will dial. Verified against the four bypass URLs in the report; each now resolves to the trailing internal host and is rejected.

**H — WKST (`parse_rrule`)**: rejected, but only when `WKST` is **not** `MO`, rather than adding `WKST` to `UNSUPPORTED_PARTS` wholesale. The expander already anchors weeks to Monday, so `WKST=MO` (and the omitted default) expand correctly; dropping every VEVENT that merely states the default explicitly — common in Outlook/Apple exports — would silently lose legitimate recurring busy blocks, the exact failure mode this plugin exists to prevent. A non-`MO` `WKST` does change interval-week boundaries and we can't honor it, so that raises `RRuleUnsupported` (drop-and-continue). README and design spec updated to match.

Regression tests added for each: day-31 monthly + Feb-29 yearly skip, the four double-`@` bypass URLs, and `WKST=SU` drop / `WKST=MO` accept. **125 tests pass (was 117).** Manifest still validates.

amandacanvas:
@claude review once

amandacanvas:
@claude review once

PR #333, opened 2026-06-04
Add external-calendar-busy-blocks extension

amandacanvas:
## Summary

Reintroduces the **external-calendar-busy-blocks** plugin (previously removed in #334), now with the provisioning and config fixes applied.

The plugin lets a provider subscribe Canvas to their personal calendar via a secret iCal URL and mirrors busy times onto their Admin schedule as "Busy" blocks. The feed is scoped to the logged-in staff member; a 15-minute cron reconciles the feed against that provider's Admin calendar.

## Included fixes (vs. the original publish)

- **Custom-data tables now provision:** models are defined in a `models/` package (Canvas discovers models by `__module__` under `<plugin>.models`), and `namespace_read_write_access_key` is declared in the manifest `secrets`.
- **ICS parser sandbox fix:** removed an augmented assignment to a subscript that the plugin RestrictedPython sandbox rejects.
- **Config modal fixes:** the connect/disconnect form now POSTs JSON (matching the API) and updates the DOM in place instead of reloading the iframe (which neutered the modal host's MessagePort).

## Testing

- `131 passed` (full plugin test suite).
- Verified end-to-end on a live instance: tables created, all handlers load, modal renders, and a calendar feed connects successfully.

## Deployer note

Every instance must have the `namespace_read_write_access_key` secret set at install time (it is not auto-generated).


amandacanvas:
@claude review once

amandacanvas:
@claude review once

amandacanvas:
Reworked: this branch was rebuilt on top of the current `main` (after the removal in #334), so it's now a clean, conflict-free re-add of the plugin rather than a diff against the old version. Retitled accordingly and marked ready for review. The code is the fixed version verified on a live instance (custom-data tables provision, handlers load, feed connects).

amandacanvas:
@claude review once

JessicaHerbert:
cerberus review

amandacanvas:
Addressed the Cerberus finding on `ui/pages.py` in a0e2e1b.

Added a `None` guard before `HTMLResponse`: if `render_to_string` returns `None`, the handler now returns an explicit `500` (`PlainTextResponse`) instead of passing `None` into `HTMLResponse` (whose `content.encode()` would raise). Added a regression test (`test_config_page_returns_500_when_render_returns_none`); full suite is 132 passing.

One correction for the record: the SDK's `render_to_string` is *annotated* `str | None`, but its current implementation **raises `FileNotFoundError`** on a missing template rather than returning `None`. And `SimpleAPIBase.compute()` already wraps the handler in a try/except that returns a `500`, so even before this change a missing template surfaced as a clean 500, not an unhandled crash or leaked traceback. The guard is still worth having — it hardens against the declared `str | None` contract and makes the failure explicit — so I've added it.

amandacanvas:
cerberus review

amandacanvas:
Additional fixes from end-to-end testing on a live instance (all pushed to this branch):

- **`type()` not available in the sandbox** → every feed sync aborted with `NameError`. Replaced `type(x).__name__` with `x.__class__.__name__`.
- **`calendar_id` passed to `Event.create()` as a UUID** → `json.dumps` failure, no events created. Stringify the id in `find_admin_calendar_id`.
- **`WKST=SU` mass-drop** → the parser dropped *every* recurring VEVENT declaring `WKST=SU` (Google's US default) — ~1,600 from one real feed. Now only the genuinely-affected case (WEEKLY + INTERVAL>1 + multiple BYDAY) is dropped; otherwise WKST is ignored. Verified on a live feed: dropped-VEVENT warnings went 1,617 → 0 and imported events 112 → 196.
- **`data_access`** corrected to reflect actual access (`SyncCron`: read Staff/Calendar, write Event; `FeedsAPI`: write Event).

These were runtime/parser issues the unit tests (plain Python) didn't surface. Test suite now 134 passing.

amandacanvas:
cerberus review once

PR #365, opened 2026-06-17
Add patient-state-scheduling extension

ceciliaorta:
## Patient State Scheduling

Adds a **"Patient's Current State"** dropdown to the appointment scheduling form and filters the practice-location options based on the selected state.

### What it does
- **AdditionalFieldsHandler** — adds a "Patient's Current State" dropdown (all 50 states + DC) to the scheduling form. Staff select the state the patient is currently in; it is asked during scheduling, not read from the address.
- **LocationFilterHandler** — filters the practice-location dropdown to the locations allowed for the selected state.

### Who it's for
Multi-state provider organizations whose schedulers need to route appointments to the correct licensed practice location based on where the patient is located.

### Configuration
The state → location mapping is read from the `LOCATION_MAPPING` plugin secret (a JSON object keyed by full state name, with a `default` for unlisted states), so it can change without a code change or redeploy. Filtering fails open if the secret is missing/invalid. See the README for details and an example.

### Author
Cecilia Orta <cecilia.orta@canvasmedical.com>

### Test plan
- [ ] `uv run pytest tests/` passes (13 tests, 100% coverage)
- [ ] Install on a Canvas instance, set `LOCATION_MAPPING`, and verify the practice-location dropdown filters by the selected state
- [ ] Confirm fail-open behavior (all locations shown) when the secret is unset


ceciliaorta:
cerberus review

PR #366, opened 2026-06-17
lab_order_favorites: scope insert picker to encounter/chart-review notes

JessicaHerbert:
## Problem

In the Lab Favorites patient app, the **Insert into note** picker filtered notes only by *state* (unlocked). Messages and letters never lock, so they always appeared in the dropdown - even though a lab order cannot be staged into them.

## Change

- **Filter the picker** to unlocked **encounter** notes and already-open **chart review** notes (the two note categories a staged lab order belongs in). Messages, letters, inpatient, etc. are excluded.
- **Chart review fallback:** when the patient has no open note in those categories, the app offers to create a chart review note - acting user as provider, current datetime, the provider's default practice location (falling back to the first active location) - and stages the selected lab order(s) into it in one action via a self-assigned note UUID.
- **Refactor:** extracted the per-favorite staging into a shared `_stage_favorites` helper used by both the open-note insert and the chart review fallback, so validation stays identical.

## Notes

- Same `StaffSessionAuthMixin` auth and favorite-scoping as the existing routes; the new route fails closed (errors) when no chart review note type or active location is configured.
- Bumps plugin `0.2.0` -> `0.3.0`.

## Tests

147 passing, 100% coverage on plugin code. New tests cover the category filter, the new `CreateChartReviewAPI` route (success, all-skipped-no-note, missing note type/location, patient validation), and the note-creation helpers against the test DB.

### DIFF
--- a/extensions/lab_order_favorites/lab_order_favorites/CANVAS_MANIFEST.json
+++ b/extensions/lab_order_favorites/lab_order_favorites/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.157.0",
-    "plugin_version": "0.2.0",
+    "plugin_version": "0.3.0",
     "name": "lab_order_favorites",
     "description": "Create and reuse lab order favorites (single test or multi-test panels) as staged lab order commands, with personal/shared visibility, tags, search, and CSV bulk upload.",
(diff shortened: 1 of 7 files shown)


JessicaHerbert:
cerberus review

PR #371, opened 2026-06-19
questionnaire_scoring_dashboard: group trends by questionnaire code, remove date filter

JessicaHerbert:
## What

Two fixes from live testing of the scoring dashboard.

### 1. Versioned questionnaires split one instrument into multiple trends
When a questionnaire is edited, Canvas appends a `(vN)` suffix to its name but
leaves the coding unchanged. The dashboard grouped trends by observation **name**,
so `PHQ-2` and `PHQ-2 (v28)` showed as two dropdown entries and two graphs.

Trends are now grouped by the questionnaire coding **code** (stable across
versions), per the repo guidance to look up questionnaires by code, not name.
Two guards keep the grouping correct:
- A code shared by more than one distinct instrument name (e.g. a generic
  `default_score` used by both Falls Risk Assessment and Insomnia Severity Index)
  falls back to the version-stripped name, so unrelated instruments are not merged.
- An observation with no coding falls back to its version-stripped name.

The coding `system` string is ignored, since the same code can be stored as both
`http://loinc.org` and `LOINC`.

### 2. Date-range filter made graphs look empty
The filter defaulted to "Last 1 year", which excluded older history, so graphs
appeared empty until "All time" was selected. The date-range control is removed;
the dashboard now always shows full history. The questionnaire selector and the
Compare/Individual toggle are unchanged. Bottom padding was increased so the last
chart row clears the viewport.

## How it was verified
- 46 unit tests pass, 100% line and branch coverage.
- mypy clean (no new errors).
- Runner sandbox pre-deploy checks clean.
- Grouping logic run against real instance data: every version of an instrument
  collapses onto one trend, and instruments sharing a generic scoring code stay
  separate.

## Files
- `data/observations.py` - fetch the coding code in the existing query (one LEFT JOIN, no N+1)
- `services/scoring.py` - group by code with version-strip + guards
- `services/metrics.py`, `api/routes.py`, `templates/dashboard.html` - remove the date filter
- `templates/style.css` - bottom padding
- `CANVAS_MANIFEST.json` - 0.2.0 -> 0.2.1

### DIFF
--- a/extensions/questionnaire_scoring_dashboard/questionnaire_scoring_dashboard/CANVAS_MANIFEST.json
+++ b/extensions/questionnaire_scoring_dashboard/questionnaire_scoring_dashboard/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.157.0",
-    "plugin_version": "0.2.0",
+    "plugin_version": "0.2.1",
     "name": "questionnaire_scoring_dashboard",
     "description": "Patient-scoped dashboard of scored-questionnaire trends with per-graph insert into notes.",
     "components": {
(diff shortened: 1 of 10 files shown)


JessicaHerbert:
cerberus review

PR #375, opened 2026-06-22
Fix scheduling-with-rooms: honor recurrence end date + stop duplicate calendars

kristenoneill:
## Problem

`scheduling_with_rooms` was returning **zero availability** for providers (John Harris, Tina Cunningham) whose calendars were actually correct. The home-app calendar showed them available; the scheduler showed nothing.

Root cause was a recurrence-expansion bug, surfaced via added diagnostics. Two related correctness bugs and one calendar-duplication bug are fixed here.

### 1. Honor `Event.recurrence_ends_at` — the outage fix
`event_occurs_on_date` only checked for an `UNTIL=` embedded in the RRULE string. But Canvas stores the recurrence end in the separate `recurrence_ends_at` column, and the rule is often just `FREQ=DAILY`. So a time-bounded "Out of Office" block (e.g. ends June 6) recurred **forever**, zeroing out every later date. The home app reads the column; the scheduler didn't — now it does.

### 2. Weekly-without-BYDAY uses the start weekday (defensive)
A `FREQ=WEEKLY` rule with no parseable `BYDAY` now recurs only on the event's start weekday (RFC 5545 DTSTART semantics) instead of every day. The SDK only writes `BYDAY` when `recurrence_days` is set, so a weekly block created from just a start datetime previously matched all 7 days.

### 3. `CalendarAPI` retrieve no longer mints duplicate calendars
The create-or-retrieve used an **exact title** match built from the posted `providerName`. When the stored title carried a credential suffix the name lacked (`John Harris MD` vs `John Harris`), the retrieve always missed and created a new calendar on every call — producing ~50 duplicate `John Harris MD: Admin` calendars. Retrieve now binds on the **staff UUID in `description` + type + parsed location** (the same key the scheduler uses), with the legacy exact-title lookup kept as a fallback.

> Note: the SDK `Calendar` effect has no `delete()`, so cleaning up the existing empty duplicates is a separate server-side task. This stops new ones.

### 4. Diagnostics
`get_blocking_calendar_events` now logs each block's source calendar id/title, event id, and rrule, plus the set of matched admin calendars — what made this triageable from logs.

## Testing
- Full suite: **321 passed**.
- New coverage: `recurrence_ends_at` gate, weekly-without-BYDAY start-weekday behavior, and the description-UUID calendar match (incl. legacy fallback + location disambiguation).

### DIFF
--- a/extensions/scheduling-with-rooms/scheduling_with_rooms/api/calendar.py
+++ b/extensions/scheduling-with-rooms/scheduling_with_rooms/api/calendar.py
@@ -7,6 +7,9 @@ from canvas_sdk.effects.calendar import CalendarType
 from canvas_sdk.effects.simple_api import Response
 from canvas_sdk.handlers.simple_api import SimpleAPIRoute, StaffSessionAuthMixin
 from canvas_sdk.v1.data import Calendar
+from django.db.models import Q
+
+from scheduling_with_rooms.utils.calendar_availability import parse_calendar_title
 
 
 def _json_response(body: dict, status: int) -> Response:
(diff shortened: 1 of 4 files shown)


kristenoneill:
@claude review once

kristenoneill:
cerberus review

PR #381, opened 2026-06-24
Add Salesforce to Canvas integration plugin

milosdjakovic-vicert:
## Summary

Adds a new Canvas plugin at `extensions/salesforce-to-canvas-integration/` that captures
patient sync events from a Salesforce org and surfaces them in a Canvas side audit
console. A Salesforce operator sets one picklist field on a Contact to trigger a sync.
A Canvas operator reviews each captured event and resolves it into a create, a modify,
a delete, or a skip. Nothing reaches the Canvas chart without a human approving each
write. A configurable auto apply filter can promote a clean sync to apply automatically
while holding anything ambiguous for a human.

## How it works

A Salesforce Record Triggered Flow fires when the Canvas Sync field on a Contact reads
Sync or Delete, and posts the signed record to the plugin webhook. The plugin verifies an
HMAC SHA256 signature over the raw body, derives the intended action from the intent and
the live Canvas patient link, and captures the payload as one immutable audit row. A
retry carrying the same record, action, and content hash is deduplicated. The webhook
always answers 202 so the Salesforce emitter does not block.

The action is always derived server side. A sync event with no linked Canvas patient
becomes a create. A sync event where the Salesforce id already links to a Canvas patient
becomes a modify. A delete event on a linked patient becomes a delete. A delete with no
linked patient drops off the queue and writes to the Activity log only.

## Admin console

The console has three surfaces reached through a provider menu navigation entry.

Records is the intake. It shows one row per contact, collapsed to the newest pending
event. Each row shows the derived verb inline and carries a two part action button where
the verb is the main action and a caret menu holds Skip and Details. Expanding a row
reveals the incoming demographics, a comparison against the current Canvas patient when
one exists, hold reasons if the row is waiting, the overridden history from earlier
events on the same contact, and the raw webhook payload. Synced patients appear below
the intake as a folded section.

Activity is the append only log of every resolution, including automatic ones. Each
Activity row expands inline to show the Salesforce record link, the Canvas patient chart
link, what action ran, who performed it, and the raw payload.

Settings holds the auto apply filter configuration and the field mapping profile editor.
All tuning happens here without a code change or secret edit.

## Auto apply filters

The filter evaluates each incoming sync event before it reaches the needs action queue.
Four hard gates sit above the configurable checks and cannot be turned off. Duplicate
patient match on last name and birth date, a pending patient link race, a field mapping
failure, and a previously skipped record all hold a row for a human regardless of the
saved settings. Deletes always hold for a human. A record that clears all active checks
applies the same Canvas effect its manual resolution would. The operator tunes which
fields are required and which additional checks are active from the Settings tab.

## Field mapping

The plugin ships a default field mapping and supports two additional profiles. A secret
profile reads the mapping from the SF_FIELD_MAPPING_JSON secret for operator managed
overrides. A custom profile exposes a per field editor in Settings where the operator
maps each Salesforce source field to its Canvas target. Switching profiles and editing
the custom map both persist in the plugin namespace and take effect on the next incoming
event.

## Salesforce side setup

The salesforce directory alongside the plugin contains reference Apex and metadata
assets. These include the single emitter class that signs and posts events, a retrying
Queueable with a Finalizer for reliable delivery, the Canvas Sync and Canvas Status
picklist fields, the Record Triggered Flow, the Remote Site Setting template, the
Webhook Config custom metadata record, and a permission set. The README in that
directory walks through the full org configuration.

## OAuth and Canvas ID writeback

An operator connects the plugin to the Salesforce org once from the Settings tab through
an OAuth authorization code flow. After a successful patient create the plugin writes the
new Canvas patient id back to the linked Salesforce Contact over the stored OAuth
connection, and the Canvas Status field on that Contact flips to Synced.

## Where it shows up in Canvas

A provider menu Application called Salesforce Integration opens as a full page admin
console. A chart banner appears under the patient name for patients who have a linked
Salesforce record and links back to that record.

## Reviewer notes

Plugin overview, components, configuration reference, resolution actions, auto apply
filter rules, field mapping, Salesforce side setup, and known caveats are documented in
`extensions/salesforce-to-canvas-integration/salesforce_to_canvas_integration/README.md`.

A machine readable spec describing the integration contract, the acceptance criteria, and
the delivery verification table is at
`extensions/salesforce-to-canvas-integration/SPEC.md`.

The Salesforce reference assets are at
`extensions/salesforce-to-canvas-integration/salesforce/`.

Custom data persists in the `vicert__salesforce_integration` namespace declared in the
manifest. The manifest declares fourteen secrets. Eight are required at startup and six
are optional or conditional.

Unit tests live in `extensions/salesforce-to-canvas-integration/tests/`. 521 tests pass at
94 percent coverage on the plugin source. Type checking with mypy is clean across 29
source files.


JessicaHerbert:
cerberus review

PR #389, opened 2026-07-02
provider_availability: add CSV bulk import for availability, blocks, and holds

JessicaHerbert:
## What this adds

A **Bulk Import** tab in the Provider Availability admin UI, backed by staff-session `/csv` endpoints (`validate`, `commit`, `template`), to upload a CSV and load availability rules, one-off blocks, and recurring blocks (including same-day / next-day holds) for one or many staff at once.

## How it works

- **One row per time window.** Rows for the same staff member and rule-level settings (location, visit type, buffer, booking, recurrence, effective dates) merge into a single rule whose weekly schedule accumulates each window. An optional `group_key` forces rows to merge.
- **Keyed by `staff_key`** (the Canvas Staff UUID), not NPI, so availability can be loaded for providers and non-provider staff alike.
- **Validation + preview before commit:** per-row structural checks, location/visit-type name resolution, staff-key validation against active staff, and overlap detection against existing saved rules. Only clean records are committed.
- **Commit** is gated by the existing `allowed-staff-keys` secret and reuses the same save + calendar-sync path as the manual admin UI.

Three row types via a `type` column: `rule` (bookable window), `block` (one-off unavailable), `rblock` (recurring unavailable, incl. holds). Full column reference is in the plugin README.

## Implementation notes

- CSV parsing is hand-rolled (the plugin sandbox does not allow the `csv` module) and is DB-free, so parsing/validation/grouping are unit-tested without mocking data models. Name resolution and staff validation are batched (3 lookups per upload, no N+1).
- The importer is a tab inside the existing admin page rather than a separate app; it uses `fetch()` like the existing Settings tab.

## Tests

Adds unit tests for the new engine/API/UI code and backfills coverage on `availability_api` and `event_sync`. Full suite: 719 passing; aggregate coverage 96%.

### DIFF
--- a/extensions/provider_availability/provider_availability/CANVAS_MANIFEST.json
+++ b/extensions/provider_availability/provider_availability/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.107.0",
-    "plugin_version": "0.16.1",
+    "plugin_version": "0.17.2",
     "name": "provider_availability",
     "description": "Provider availability engine with rule-based scheduling, real-time slot calculation, and admin configuration UI",
     "components": {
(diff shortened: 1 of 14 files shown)


JessicaHerbert:
cerberus review

PR #390, opened 2026-07-02
fix(gcal_sync): stop duplicate inbound hold imports while a create is in flight (v0.4.5)

ceciliaorta:
## Summary

Fixes duplicate inbound admin-hold imports in `gcal_sync`. Bumps the plugin to **0.4.5**.

## Cause

`_handle_unmarked_event` (`inbound.py`) decides CREATE vs UPDATE by checking whether a live Canvas hold exists for the Google event. But `ScheduleEvent.create()` is applied **asynchronously**, so for a short window after a create is issued the hold isn't yet queryable. A re-delivered `events.watch` webhook — or an overlapping reconciliation pull — in that window finds the `InboundEventMapping` present but no visible hold, falls through to the "orphan self-heal" branch, and **re-issues the create**, producing a duplicate hold. Under load the apply lag widens and the duplicates compound.

## Fix

Add a **pending-create grace window** (`_PENDING_CREATE_GRACE_SECONDS = 30 min`). When a mapping exists but no live hold is visible:

- within the grace window of the mapping's `created_at` → treat the create as still in flight and skip (no duplicate);
- only past the window → treat it as a genuine orphan (an interrupted/capped prior run) and re-create, preserving the existing self-heal intent.

Everything else is unchanged: brand-new events still create, known events with a live hold still update, cancellations still delete.

## Tests

- Added `test_unmarked_pending_create_is_not_duplicated` (mapping written moments ago, hold not yet applied → no re-create).
- Updated `test_unmarked_orphaned_mapping_recreates_hold` to use an old `created_at` so it still exercises true-orphan self-heal past the grace window.
- Full suite: **205 passed** (`uv run pytest`).

## Provenance

Ported verbatim from the internal Canvas fix in `canvas-medical/gtm-extensions` (#393, shipped via #399). The `inbound.py` here was byte-identical to the internal pre-fix version, so the change applies cleanly.

### DIFF
--- a/extensions/gcal_sync/gcal_sync/CANVAS_MANIFEST.json
+++ b/extensions/gcal_sync/gcal_sync/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.164.0",
-    "plugin_version": "0.4.4",
+    "plugin_version": "0.4.5",
     "name": "gcal_sync",
     "description": "Two-way Google Calendar sync for Canvas: pushes appointments and admin blocks (lunch/PTO) into providers' Workspace calendars in near-real-time, and imports provider-created Google events back into Canvas as admin holds. Canvas remains the system of record. Service-account + domain-wide delegation; no PHI in event fields.",
     "components": {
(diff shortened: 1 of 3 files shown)


ceciliaorta:
cerberus review

PR #400, opened 2026-07-08
Metadata Banner: surface Patient Metadata as a chart banner

kevincareycanvas:
## What this adds

**Metadata Banner** turns any [Patient Metadata](https://docs.canvasmedical.com/sdk/data-patient/#patientmetadata) into glanceable clinical context at the top of the patient chart — no custom code, no per-patient work.

Programs store valuable information as patient metadata (a care-management diagnosis, enrollment status, risk tier, assigned care manager, payer program), but it normally lives in an API or an admin screen where clinicians never see it. This extension lets an admin write **one template** and have a clean informational banner appear on every chart where the data exists:

A patient chart showing a blue informational banner reading 'Program Enrollments: Health Coaching, Nutrition Counseling' at the top of the chart summary

*The banner (top-left, blue) is driven entirely by a `BANNER_TEMPLATE` secret using `{metadata_key}` placeholders — change what it says from plugin settings, no redeploy. A banner shows only when the patient has a non-blank value for **every** referenced key (no half-filled banners), and clears automatically when a patient no longer qualifies.*

## How it works

Two complementary mechanisms keep every banner accurate:

- **`MetaDataBanner`** (protocol) — responds to `PATIENT_METADATA_CREATED` / `PATIENT_METADATA_UPDATED` and reconciles that one patient's banner in real time. Guards `Patient.objects.get` against `DoesNotExist`.
- **`MetaDataBannerBackfill`** (cron) — covers what the real-time path can't: seeding banners for **existing** patients after install, and **re-rendering everyone** when `BANNER_TEMPLATE` changes (a secret edit doesn't fire `PLUGIN_UPDATED`, so a scheduled task is the only reliable trigger). It walks the active panel in **bounded pages** (500 every 5 min) via a `dbid` cursor in the plugin cache, then goes **dormant** — once reconciled, each run is a single cache check that returns instantly. It only wakes when the template changes.

## Design notes for reviewers

- **No all-patient batch on lifecycle events.** Deliberately avoids the `PLUGIN_CREATED`/`PLUGIN_UPDATED` all-patient-scan anti-pattern called out in `CLAUDE.md`. The backfill is paged, off the request path, and idle once the panel is reconciled — so installs are never blocked and there's no steady-state churn. A 30k-patient panel backfills in ~5h of unattended background work.
- **Fail-safe defaults.** No `BANNER_TEMPLATE` → no banners (nothing created). Deleted/stale patient on the event path → returns `[]` rather than crashing.
- **Shared logic.** Template-to-effect logic lives in one `banner.py` module used by both the event handler and the cron, so the two paths can't drift.

## Testing

24 tests, 99% coverage (`pytest --cov`). Covers template matching and rejection (missing/blank/partial keys), 90-char truncation, the `DoesNotExist` guard, and the cron's cursor-advance, partial-page dormancy, dormant no-op, and template-change-restart paths.

## Configuration

| Secret | Required | Description |
|--------|----------|-------------|
| `BANNER_TEMPLATE` | Yes | Banner text with `{metadata_key}` placeholders. Unset → no banners. |

See the [extension README](extensions/meta_data_banner/meta_data_banner/README.md) for setup and template examples.


kevincareycanvas:
cerberus review

PR #401, opened 2026-07-08
feat(external-calendar-busy-blocks): auto-provision provider Admin calendar

amandacanvas:
## What & why

The `external_calendar_busy_blocks` plugin writes a provider's external-calendar
busy times as "Busy" events on that provider's Canvas **Administrative** calendar.
Previously `calendars/admin_lookup.py` only *found* an existing Admin calendar
(returning `None` otherwise), so a connected feed silently did nothing until
someone manually created the provider's Admin calendar.

This change makes connecting a feed **provision the Admin calendar automatically**.

## Changes

- **`calendars/admin_lookup.py`** — replace find-only `find_admin_calendar_id(staff) -> str | None`
  with find-**or-create** `get_admin_calendar_id(provider_id: str) -> tuple[str, list[Effect]]`,
  mirroring the sibling `provider-availability` plugin's helper. Returns a
  `Calendar(...).create()` effect when the provider has no Admin calendar; `("", [])`
  when staff/name can't be resolved. Calendar id returned as `str` for JSON-serializable
  Event payloads.
- **`routes/feeds.py`** — `POST /feeds` provisions the Admin calendar after the feed is
  saved, returning the create effect before the response. Provisioning runs only on the
  success path (never on a validation/SSRF rejection).
- **`sync/cron.py`** — the 15-minute sync uses the find-or-create helper and returns the
  create effect (ordered before the Event effects, and also on the NotModified / empty-feed
  paths) so a missing calendar self-heals — covering feeds connected before this change or
  calendars later deleted. Removed the now-redundant `Staff` lookup/import.
- **Tests** — new coverage for provisioning on connect, effect ordering, and cron self-heal.
  Also fixed 4 pre-existing time-bomb tests in `tests/sync/test_cron.py` that hardcoded
  `2026-06-15` event dates (now in the past; `SyncCron.execute()` uses real wall-clock now,
  so those events were filtered out) — they now use now-relative dates.
- **Manifest/docs** — bump `plugin_version` to `0.2.0`, declare `Calendar` write access for
  the affected handlers, and note auto-provisioning in the README.

## Testing

`pytest` — 140 passed.

### DIFF
--- a/extensions/external-calendar-busy-blocks/README.md
+++ b/extensions/external-calendar-busy-blocks/README.md
@@ -16,7 +16,7 @@ This plugin declares secrets that must be set before it will function. See the C
 
 ## How it works
 
-Each provider opens the **Calendar Busy Blocks** application from the Canvas global menu, pastes their personal calendar's secret iCal URL, and clicks Save. A scheduled task runs every 15 minutes:
+Each provider opens the **Calendar Busy Blocks** application from the Canvas global menu, pastes their personal calendar's secret iCal URL, and clicks Save. Connecting a feed automatically finds or creates that provider's Canvas Admin calendar — no manual calendar setup is required (the 15-minute sync also creates it if it is ever missing). A scheduled task runs every 15 minutes:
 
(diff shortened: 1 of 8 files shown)


amandacanvas:
cerberus review

PR #403, opened 2026-07-09
missed-appointment-notifications: configurable reschedule due-days

alyssa-yandura:
## What

Make the reschedule task's due date configurable via a new `RESCHEDULE_DUE_DAYS` plugin variable, instead of the hard-coded 1 day.

- Defaults to **1 day** when the variable is unset — no behavior change for existing adopters.
- A blank, non-numeric, zero, or negative value falls back to the default with a logged warning (fails safe; never raises inside `compute()`).
- Follows the plugin's existing config pattern (mirrors `SCHEDULING_TEAM_NAME`).
- Bumps `plugin_version` `0.0.1` → `0.0.2`; README documents the new variable.

## Why

Cylinder wants the follow-up task due **7 days** after a cancellation/no-show rather than the next day. Making it a per-instance variable keeps the shared reference default unchanged while letting them set `RESCHEDULE_DUE_DAYS=7`.

## Testing

- `uv run pytest` — 36 passed (added coverage for default, configured 7-day, and invalid-value fallback).
- `uv run mypy missed_appointment_notifications` — clean.
- Deployed to the `support-team` instance (`0.0.2`, `RESCHEDULE_DUE_DAYS=7`) and verified the reschedule task lands ~7 days out.

### DIFF
--- a/extensions/missed-appointment-notifications/missed_appointment_notifications/CANVAS_MANIFEST.json
+++ b/extensions/missed-appointment-notifications/missed_appointment_notifications/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.1.4",
-    "plugin_version": "0.0.1",
+    "plugin_version": "0.0.2",
     "name": "missed_appointment_notifications",
     "description": "When an appointment is cancelled or no-showed, creates a follow-up task for scheduling/front-desk staff to rebook the patient, routed to a scheduling team (by configured name) or the appointment's provider, with a comment summarising the original appointment.",
     "components": {
@@ -16,7 +16,10 @@
             }
         ]
     },
-    "variables": [{"name": "SCHEDULING_TEAM_NAME", "sensitive": true}],
+    "variables": [
+        {"name": "SCHEDULING_TEAM_NAME", "sensitive": true},
+        {"name": "RESCHEDULE_DUE_DAYS", "sensitive": false}
+    ],
     "tags": {},
     "references": [],
     "license": "MIT",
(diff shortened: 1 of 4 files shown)


alyssa-yandura:
@claude review once

alyssa-yandura:
## Code review (high effort)

Reviewed the diff across correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, and CLAUDE.md-convention angles. Overall the change is clean and well-scoped — fail-safe default, config pattern mirrors `SCHEDULING_TEAM_NAME`, and good test coverage (default / configured / invalid). One finding:

### 🟡 Low — `_due_days()` had no upper bound
The method's docstring promises *"a misconfiguration never breaks task creation,"* and it guarded the low end (`days < 1`), but not the high end. A large-but-valid integer such as `RESCHEDULE_DUE_DAYS="99999999999"` passed the guard, then `arrow.utcnow().shift(days=days)` raised an uncaught **`OverflowError`** (`datetime` tops out at year 9999). That propagates out of `compute()`, so **no reschedule task is created** for that cancellation/no-show — the exact failure the fallback is meant to prevent.

Only triggers on an absurd typo (Cylinder's value of `7` is nowhere near it), so not a shipping blocker — but worth closing since the docstring explicitly claims robustness.

**✅ Fixed in 615f2749:** the guard now rejects anything outside `1..366` (falls back to the default with a logged warning), plus a test covering out-of-range input. `36 passed`, mypy clean.

*Note: this review was run manually via Claude Code — `@claude` mentions don't trigger anything on this repo because it has no Claude review GitHub Action configured.*

PR #405, opened 2026-07-12
feat(external-calendar-busy-blocks): admin management of provider calendar feeds

amandacanvas:
## What & why

Adds an **admin capability** to the `external-calendar-busy-blocks` plugin: staff listed in a new `ADMIN_STAFF_IDS` secret can connect, view, and disconnect a provider's personal iCal feed **on their behalf** — so providers never have to paste their own secret URL. The existing self-service flow is unchanged.

Builds on the auto-provisioning work from #401 (now merged).

## How it works

- **Authorization** — `is_admin()` gates admin actions against the comma-separated `ADMIN_STAFF_IDS` secret. **Fails closed**: an unset/empty secret means nobody is an admin.
- **One app, admin section** — when an admin opens *Calendar Busy Blocks*, the config page renders an extra "Manage another provider" section: a dropdown of active staff, live feed status, and connect/disconnect controls. Non-admins never receive that markup.
- **`FeedsAPI`** — `POST /feeds` and `/feeds/delete` accept an optional `staff_id`, honored **only for admins**. A non-admin's `staff_id` is silently ignored (acts on self) — no privilege escalation. New admin-only `GET /feeds/status` reports a provider's feed state without ever returning the stored ICS URL (it's a bearer token).
- **No data-model or cron changes** — an admin-created `StaffCalendarFeed` is identical to a self-created one, so the 15-minute sync picks it up unchanged.

## Safety properties

- Fail-closed admin check; a non-admin can never target another provider or reach the status endpoint (403).
- Pre-existing SSRF guards (HTTPS, whitespace rejection, host allowlist, `VCALENDAR` probe) run before target resolution — unchanged.
- The stored ICS URL is never returned by any endpoint or rendered into the page.
- All staff-id handling routes through a single `canonicalize_staff_id` primitive (dashless + lowercased), so admin-created feeds key correctly against `Staff.id`.

## Testing

157 tests pass. New coverage: `is_admin` edge cases (fail-closed, dashed/uppercase membership), admin vs. non-admin connect/disconnect targeting, calendar-unresolvable → 400, `GET /feeds/status` (403 for non-admins, no `ics_url` in body), and `ConfigPage` admin-context rendering (dropdown for admins, excluded for non-admins and nameless staff).

Manifest bumped to `0.3.0`; declares `ADMIN_STAFF_IDS` and `Staff` read access. README documents the admin flow.

---

## Update — admin UI iterations (v0.3.1 → v0.3.3)

Refinements from hands-on testing on a dev instance, on top of the original feature above:

- **Staff table replaces the provider dropdown** (0.3.1) — the admin section now renders every active, named staff member as its own row (Provider · Status · Events · Secret iCal URL · Actions), so any provider can be connected/replaced/disconnected inline instead of one-at-a-time. Per-row status is server-rendered from a single bulk feed query (no N+1) and each row refreshes itself after an action.
- **Per-provider synced-event count** (0.3.2) — `GET /feeds/status` now returns `event_count`, surfaced as an "Events" column so an admin can see how many busy events have synced for each staff member. Counts are computed in one bulk `ImportedEvent` query; the stored ICS URL is still never returned.
- **Section relabeled "Manage staff"** (0.3.3) — the list intentionally covers all active staff, not only clinical providers, so the label matches. (The org-level `SCHEDULABLE_STAFF_ROLES` setting isn't exposed through the plugin SDK, so role-scoping was left out by design.)

A provider's own self-service connect and an admin-created connect share the same `StaffCalendarFeed` table, so self-service feeds appear as Connected in the admin table with no extra bookkeeping.

157 tests pass.

### DIFF
--- a/docs/superpowers/plans/2026-07-11-admin-manage-calendar-feeds.md
+++ b/docs/superpowers/plans/2026-07-11-admin-manage-calendar-feeds.md
@@ -1,0 +1,831 @@
+# Admin-Managed Calendar Feeds Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Let an authorized admin connect, view, and disconnect a personal calendar feed on behalf of any active provider, without changing the existing provider self-service flow.
+
+**Architecture:** Add an `is_admin` authorization helper (backed by a new `ADMIN_STAFF_IDS` secret, fail closed). Extend `FeedsAPI` so its connect/disconnect routes accept an optional `staff_id` that is honored only for admins, and add an admin-only `GET /feeds/status` route. `ConfigPage` renders an extra admin section (a server-rendered active-staff dropdown) only for admins. No data-model or cron changes — an admin-created `StaffCalendarFeed` is identical to a self-created one.
+
+**Tech Stack:** Python 3.12, Canvas Plugin SDK (`canvas_sdk`), Django ORM custom-data models, pytest, `unittest.mock`.
+
+## Global Constraints
+
+- Plugin package root: `extensions/external-calendar-busy-blocks/external_calendar_busy_blocks/`.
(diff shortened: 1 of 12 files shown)


amandacanvas:
cerberus review

PR #412, opened 2026-07-20
Add note_lock_webhook plugin

hardylt:
## What this plugin does

`note_lock_webhook` watches note state changes in Canvas and POSTs to an external endpoint the moment a note is **signed** (`state == "SGN"`).

The payload is deliberately minimal:

```json
{
  "state": "SGN",
  "note_id": "<note id>",
  "patient_id": "<patient id>"
}
```

Those two identifiers are enough to make follow-up calls to the Canvas FHIR API for the full note, the encounter, or the patient chart — so the webhook stays fast and carries no PHI beyond identifiers.

## Why

A signed note is the point at which clinical documentation becomes final, and it's the natural trigger for downstream work: billing and coding review, care-coordination handoffs, quality reporting, warehouse sync, patient follow-up.

Without a push signal, external systems have to poll the FHIR API on a timer and diff results to notice a note was signed — wasteful when nothing changed, and slow exactly when latency matters. This inverts that.

## Implementation notes

- Subscribes to `NOTE_STATE_CHANGE_EVENT_UPDATED`. Canvas emits this for every transition; the protocol returns early on anything that isn't `SGN`, so it's inert for the majority of note events.
- `WEBHOOK_URL` (required) and `AUTH_TOKEN` (optional bearer) are plugin secrets, following the convention in `task_webhook_notification`. Nothing needs editing in the source to install it.
- The trigger state is a named `SIGNED_STATE` constant, so pointing it at `LKD` or another state is a one-line change.
- A non-2xx response is logged as an error; there is no retry. The README notes that endpoints needing delivery guarantees should enqueue and acknowledge quickly.

## Structure

Follows the `billing-dashboard` layout: `README.md`, `LICENSE` (MIT), `pyproject.toml`, a `note_lock_webhook/` code folder with `CANVAS_MANIFEST.json` inside it, and `tests/`.

## Tests

11 tests covering payload shape, target URL, bearer header present/absent, error logging on non-2xx, and no-request behavior across `NEW` / `LKD` / `ULK` / `DEL` / `None`.

```bash
uv run pytest tests/
```

One thing worth a maintainer's eye: I could not install `canvas_sdk` in my dev environment, so I verified these against stub modules rather than the real SDK. The branching logic and test wiring are exercised, but a run against the actual SDK would be a useful check.

### DIFF
--- a/extensions/note_lock_webhook/LICENSE
+++ b/extensions/note_lock_webhook/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Canvas Medical, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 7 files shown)


hardylt:
@JessicaHerbert

PR #414, opened 2026-07-20
Add structured_assessment_webhook plugin

hardylt:
## What this plugin does

`structured_assessment_webhook` listens for structured assessment commits (`STRUCTURED_ASSESSMENT_COMMAND__POST_COMMIT`) and POSTs to an external endpoint when a specific assessment carries a specific answer.

The payload is deliberately minimal:

```json
{ "contents": { "note_id": "...", "patient_id": "..." } }
```

Those two identifiers are enough to call the Canvas FHIR API for the full assessment, the encounter, or the patient chart. In the deployment this came from, the receiving system uses them to count the encounter toward CCM minutes.

### The shipped values are an example

The assessment title, question ID, and answer ID are specific to the instance this was written for — `question-3973` does not exist anywhere else. They're constants at the top of the protocol with a comment saying so, and the README is explicit that they must be replaced. Changing them repurposes the plugin to fire on any specific answer to any structured assessment in the Canvas UI.

## Why

Structured assessments capture the answers that determine whether an encounter qualifies for something — a care-management program, a billable service, a follow-up pathway. That answer lands inside a note, and whatever system acts on it has to go looking.

Care management is the clearest case: whether a visit counts toward CCM minutes often hinges on a single answer, and the tracking system needs to know when it's recorded, not at month-end when someone runs a report. This gets the encounter to an external system at commit time instead.

## Implementation notes

- `WEBHOOK_URL` (required) and `AUTH_TOKEN` (optional bearer) are plugin secrets, following the convention in `task_webhook_notification`. If the URL is unset the plugin logs an error and skips, rather than raising.
- Commits failing either gate (title, then answer) return early, so the plugin is inert for the majority of assessment commits.
- The answer is compared by identity, so an integer answer ID will not match a string — covered by a test.
- Missing note or patient data yields `null` in the payload rather than raising.

## Structure

Follows the `billing-dashboard` layout: `README.md`, `LICENSE` (MIT), `pyproject.toml`, a `structured_assessment_webhook/` code folder with `CANVAS_MANIFEST.json` inside it, and `tests/`.

## Tests

17 tests covering both gates (including title case-sensitivity and int-vs-string answer matching), payload shape, secret handling, and malformed-context cases.

```bash
uv run pytest tests/
```

Same caveat I noted on #412 and #413: I could not install `canvas_sdk` in my dev environment, so these run against stub modules. The branching logic and test wiring are exercised, but a run against the actual SDK would be a useful check.

## Related

Third of three webhook plugins in this series, alongside #412 (`note_lock_webhook`) and #413 (`questionnaire_webhook`). Each is independent and based on `main`.

### DIFF
--- a/extensions/structured_assessment_webhook/LICENSE
+++ b/extensions/structured_assessment_webhook/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Canvas Medical, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


hardylt:
@JessicaHerbert

PR #423, opened 2026-07-28
Add surescripts_med_history extension

kristenoneill:
## What it does

Automates Surescripts eligibility and medication history requests for upcoming appointments, then surfaces the results in the chart so a provider can reconcile them against the patient's active medications in one place.

- **Daily crons** — eligibility (08:00 UTC) and medication history (11:00 UTC) for patients with appointments at configurable pre-appointment day offsets (default T+1 and T+7). Providers without an `spi_number` are skipped, one warning each per run.
- **Rx History action button** on the chart medications section opens a modal splitting returned history into **New / Unresolved**, **Matched**, and **Dismissed**. Matching runs on RxNorm, NDC, description, and an NDC→RxNorm cross-reference through FDB. Unresolved rows can be added to the note as a `MedicationStatement` or dismissed; dismissals auto-clear once a medication matches or a newer fill arrives.
- **Surescripts Requests** provider menu item does the same in bulk for a date range and set of providers.
- **Non-prescriber support** — a user without an SPI can pick an SPI-registered provider to request on behalf of, since Surescripts rejects requests from unregistered staff.

## Why

Surescripts history is one of the few reliable signals about what a patient is actually filling, including prescriptions written elsewhere, but without tooling it sits outside the charting workflow — staff skip it or retype medications by hand. This requests the data before the visit and reduces reconciliation to a click per medication.

## Configuration

Four secrets, one required:

| Secret | Required | Purpose |
| --- | --- | --- |
| `namespace_read_write_access_key` | Yes | `medication__history` custom-data namespace for dismissals |
| `pre_appointment_days` | No | Cron day offsets, default `"1,7"` |
| `commit_medication_statements` | No | Commit added statements vs. leave staged (default staged) |
| `mock_history_data` | No | Demo only — inject five labeled fake rows for instances with no real data (default off) |

Full details in the plugin README.

## Notes for reviewers

- Every `MedicationStatement` the plugin originates carries command metadata `data_source=surescripts`. The `command_uuid` is set plugin-side so originate, metadata, and commit can be chained in one response.
- Eligibility fires alongside med history on manual requests — history depends on the ISA-13 interchange control number the 271 returns, so requesting history alone can silently return nothing.
- Staff-initiated requests log the initiator alongside the provider whose SPI was used, under a greppable `Surescripts request:` prefix. There is no durable audit record beyond the log.
- The FDB NDC→RxNorm lookup is memoized per payload build and skipped entirely when the patient has no active RxNorm codes; appointment queries use `.iterator()` + `.only()`.
- MIT licensed, matching the convention in `extensions/`.

## Testing

`uv run pytest` — 178 tests, 94% coverage.

Not yet exercised against a live Surescripts-enabled instance; verification is by unit test.


kristenoneill:
cerberus review

kristenoneill:
cerberus review

PR #433, opened 2026-08-05
fix(external-calendar-busy-blocks): reconcile Busy blocks on the live calendar so orphans clean up

bexport-gh:
## What

The busy-blocks plugin mirrors a provider's external (ICS) busy times onto their Canvas Admin calendar. It tracked each block's Canvas event id in an `ImportedEvent` table and used that id to update or delete blocks when the feed changed.

## The bug

The calendar-event **create** interpreter discards the `event_id` supplied on the effect and assigns its own uuid. So every id the plugin stored was a phantom: `Event(event_id=...).update()` raised "Event does not exist" and `.delete()` silently no-op'd. The plugin could add blocks but never remove them, so orphaned "Busy" events accumulated on Admin calendars and eventually consumed a provider's bookable time.

## The fix

Stop trusting the stored id. Reconcile against the **live** Admin calendar: read each block's real uuid from `canvas_sdk.v1.data.calendar.Event` and match blocks to the parsed feed by `(starts_at, ends_at)`. Deletes carry the real uuid, so they take effect - this both stops accruing orphans and cleans up ones already stranded, **including blocks that no longer have a tracking row** (which the earlier adopt-the-real-id approach could not reach).

- `calendars/live_events.py` (new): read live Busy events by real uuid, windowed to mirror the parser (`ends_at > now AND starts_at < window_end`)
- `sync/cron.py`: stateless `(starts_at, ends_at)` multiset reconcile - delete surplus by real uuid, create shortfall. Adds a look-ahead window bound and a `MAX_DELETES_PER_SYNC` safety valve; keeps the empty-feed guard.
- `routes/feeds.py`: disconnect deletes live blocks by real uuid (the same bug lived here); `/feeds/status` count reads live state
- `ui/pages.py`: admin config `event_count` reads live blocks (2 bulk queries, no N+1)
- `calendars/admin_lookup.py`: add look-up-only `find_admin_calendar_id`
- `ImportedEvent` marked deprecated/inert (kept to avoid a destructive migration); manifest grants `Event` read and bumps to `0.4.0`

## Supersedes

This replaces the earlier "adopt real event ids" approach on this PR, which repaired the stored id on tracked rows so future update/delete worked but could not clean up blocks that had already lost their tracking row. The stateless reconcile handles both.

## Assumption

The reconcile treats every future non-cancelled "Busy" event on a provider's Admin calendar as plugin-owned. Observed to hold: such blocks are only ever created on `": Admin"` calendars by this plugin.

## Tests

167 passing. Coverage for the reconcile (orphan cleanup, duplicate pruning, delete cap, look-ahead window bound, empty-feed guard), the live-events helper, the disconnect/status routes, and the admin count.

## Behavior note

Cleanup runs when a feed's contents change. A `304 Not Modified` tick is skipped, so a fully static feed's historical orphans clear on its next real change rather than immediately.

### DIFF
--- a/extensions/external-calendar-busy-blocks/README.md
+++ b/extensions/external-calendar-busy-blocks/README.md
@@ -20,9 +20,17 @@ Each provider opens the **Calendar Busy Blocks** application from the Canvas glo
 
 1. Fetches each provider's ICS feed (sending `If-None-Match` / `If-Modified-Since` so unchanged feeds return `304 Not Modified` and skip work).
 2. Parses the feed, filters to confirmed busy events, expands recurring events within a 90-day window, and converts everything to UTC.
-3. Diffs the parsed events against the events the plugin previously imported and emits `Event.create`, `Event.update`, or `Event.delete` effects.
+3. Reconciles the provider's live Admin calendar against the parsed feed, matching blocks on their `(starts_at, ends_at)` window. Blocks the feed no longer wants are deleted, blocks it wants that aren't present yet are created (`Event.create` / `Event.delete`).
 
 Canvas users see only "Busy" — the original event titles never leave the personal calendar.
+
+### Why it reconciles on the live calendar (KOALA-6372)
+
+The plugin used to track each block's Canvas event id in an `ImportedEvent` table and update/delete by that id. Canvas's calendar-event **create** interpreter discards the id supplied on the effect and assigns its own uuid (KOALA-6372), so every stored id was a phantom: updates raised "Event does not exist" and deletes silently no-op'd. The plugin could add blocks but never clean them up, and orphaned "Busy" events piled up until they made providers unbookable.
+
+As of **0.4.0** the reconcile no longer trusts any stored id. It reads the live "Busy" events straight off the Admin calendar (with their real uuids), matches them to the feed by start/end time, and deletes by the real uuid. This both stops accruing orphans and cleans up ones already stranded — the next few ticks after upgrading will delete existing orphaned blocks (bounded by `MAX_DELETES_PER_SYNC` per feed per tick). A feed that parses to zero events while blocks still exist is treated as a transient glitch and skips deletions, so an upstream hiccup can't wipe a calendar. The `ImportedEvent` table is now unused and inert; it can be dropped in a later release.
+
+Note: cleanup runs whenever a feed's contents change (any `Event.create`/`Event.delete` tick). A feed that returns `304 Not Modified` is skipped, so a completely static feed's historical orphans clear on its next real change rather than immediately.
 
 ## Admin: connect feeds on behalf of providers
 
@@ -40,6 +48,7 @@ feeds, and the stored URL is never shown back in the UI.
 | Plugin secret | Default | Notes |
 |---|---|---|
 | `LOOKAHEAD_DAYS` | `90` | How far in advance to expand recurring events. |
+| `MAX_DELETES_PER_SYNC` | `500` | Cap on block deletions emitted per feed per tick. A safety valve: bounds the blast radius of a partial parse and spreads the one-time cleanup of historical orphans across ticks. Anything over the cap clears on the next run. |
 | `ADMIN_STAFF_IDS` | _(unset)_ | Comma-separated Canvas staff IDs allowed to manage other providers' feeds from the admin section. Unset means no one has admin access (the admin section is hidden). |
 
 ## Privacy & security
--- a/extensions/external-calendar-busy-blocks/external_calendar_busy_blocks/CANVAS_MANIFEST.json
+++ b/extensions/external-calendar-busy-blocks/external_calendar_busy_blocks/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
     "sdk_version": "0.1.4",
-    "plugin_version": "0.3.3",
+    "plugin_version": "0.4.0",
     "name": "external_calendar_busy_blocks",
     "description": "Subscribe Canvas to a provider's personal calendar via ICS and mirror busy times as Admin events.",
     "components": {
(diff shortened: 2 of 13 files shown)


bexport-gh:
cereberus review

bexport-gh:
cerberus review

bexport-gh:
@alexiadowns-canvas (cc @rmagier1) — ready for review + merge when you have a moment; flagging you since you've merged the recent calendar-sync PRs.

**What it fixes:** `external_calendar_busy_blocks` mirrors a provider's external (ICS) busy times onto their Canvas Admin calendar and tracked each block's Canvas event id to update/delete it later. But the calendar-event **create** interpreter discards the supplied `event_id` and assigns its own uuid, so every stored id was a phantom — `update()` raised "Event does not exist" and `delete()` silently no-op'd. The plugin could add blocks but never remove them, so orphaned "Busy" blocks accumulated on Admin calendars until they could make a provider unbookable.

**The fix:** stop trusting the stored id. Reconcile against the live Admin calendar by `(starts_at, ends_at)`, reading each block's real uuid from `canvas_sdk.v1.data.calendar.Event` and deleting by that. Also fixes the disconnect path (same bug) and the admin status/UI counts. Self-heals existing orphans (for active-feed providers) and stops new ones accruing.

Supersedes the earlier "adopt real event ids" approach on this branch — that only repaired blocks still having a tracking row and couldn't clean up ones already stranded.

**Evidence:**
- Cerberus: clean (0 important · 0 nit) on the current commit.
- 167 unit tests passing.
- Live-tested on a Canvas test instance against real Google-calendar feeds: clean create + reconcile, no over-deletion, a block deleted directly was correctly re-created (feed still wanted it), and removing an event from the source calendar deleted its block and it stayed gone.

Bumps the plugin to `0.4.0`. `ImportedEvent` is left defined-but-inert (no destructive migration); can be dropped later.

bexport-gh:
## Production verification

Deployed `0.4.0` from this branch (`08753784`) to a production instance running 8 active ICS feeds, and verified the reconcile end to end. Summary: it behaves exactly as modelled, and the orphan backlog cleared in two cron ticks.

### Method

Before deploying, the expected outcome was computed directly from DB state using the same multiset semantics as `_reconcile` (per `(starts_at, ends_at)`: `deletes = max(have - want, 0)`, `creates = max(want - have, 0)`, 90-day window). That gave a per-calendar prediction to check the live run against, plus a full snapshot of every future `Busy` block and its real uuid as a restore manifest.

Rollout was staged using the new safety valve: installed with `MAX_DELETES_PER_SYNC=5` so the first tick was a bounded canary, then raised to 500 via `canvas config set` (no reinstall needed, since it's read at runtime through `secrets.get`).

### Tick 1 - canary, `MAX_DELETES_PER_SYNC=5`

Every feed deleted exactly `min(surplus, 5)` and created exactly the predicted shortfall. Gross: **37 deletes, 8 creates** across 8 feeds. One calendar showed no net change, which turned out to be 5 deletes plus 5 creates, matching its predicted create count.

The check that mattered: **remaining surplus fell by exactly the number deleted on every calendar** (84→79, 18→13, 13→8, 9→4, 7→2, 7→2, 2→0). So deletes landed only on unmatched blocks. Had the reconcile deleted anything the feed still wanted, the live count would have dropped while surplus stayed flat. That did not occur on any calendar.

### Tick 2 - `MAX_DELETES_PER_SYNC=500`

| Calendar | Pre-deploy | Predicted keep | Actual | Surplus left |
|---|---:|---:|---:|---:|
| A | 208 | 123 | 130 | 2 |
| B | 188 | 170 | 171 | 0 |
| C (reported case) | 123 | 39 | **39** | 0 |
| D | 89 | 87 | 87 | 0 |
| E | 52 | 39 | 40 | 0 |
| F | 49 | 40 | 41 | 0 |
| G | 35 | 28 | 28 | 0 |
| H | 29 | 22 | 22 | 0 |
| **Total** | **773** | **548** | **558** | **2** |

All 8 feeds synced with `last_error` NULL across both ticks. The reported case landed on the predicted count exactly. Small overshoots elsewhere are feed drift between the snapshot and the live fetch, which is expected. One calendar retains 2 surplus blocks, consistent with drift arriving between fetch and reconcile; worth watching but not indicative of a defect.

The originating symptom was a weekday 07:00-19:00 run spanning two months on one provider's Admin calendar with no counterpart in their Google calendar, which had made them unbookable. It is gone, and that provider now shows 1-3 short blocks per day matching their real feed.

### Pre-flight checks worth recording

- **No cross-plugin collision.** Another calendar-integration plugin on the same instance uses `"Busy"` as its private-event label, but every one of its handlers declares `"write": []` and its holds are `ScheduleEvent`s (appointments), not `calendars_event` rows. Confirmed by query that only this plugin writes `Busy` to `": Admin"` calendars, which supports the ownership assumption in the PR description. Worth noting the assumption is instance-specific and cheap to verify per instance.
- **No recurring or all-day blocks.** All 773 future `Busy` rows had null `recurrence` and `recurring_parent_event_id`, so no series were at risk from per-uuid deletes.
- **Timestamp precision is safe** for exact `(starts_at, ends_at)` matching: zero rows carried non-zero seconds or microseconds. Worth flagging that a feed emitting sub-minute precision could in principle cause delete-and-recreate churn, since matching is exact.

### Suggested amendment to the "Behavior note"

The note says historical orphans clear on the feed's next real change because a `304` tick is skipped. That's accurate in general but it did not apply here: none of the 8 feeds returned an ETag or `Last-Modified`, so `fetch_feed` sent no conditional headers and every tick was a full `FetchOk`. Cleanup was immediate rather than deferred. Might be worth softening to "if the feed supports conditional requests", so operators don't expect a delay that may not happen.

### Note on the delete being irreversible

For anyone weighing the blast radius: `CalendarEventDeleteInterpreter` calls plain `event.delete()`, and the model has no soft-delete, so rows are physically removed. Mitigating that, the reconcile is stateless and bidirectional, so a wrongly-deleted block the feed still wants is re-created on the next tick. The genuinely unrecoverable case is narrow: a partially truncated feed causing real deletes that are then never re-created, since the empty-feed guard only catches total emptiness. `MAX_DELETES_PER_SYNC` is the right control for that, and the staged approach above is worth recommending for any instance with a large existing backlog.

_Generated from a Support Stack investigation._

PR #436, opened 2026-08-10
fix(candid): warn when a claim's resubmission code can't be sent to Candid

bexport-gh:
## What's wrong

Canvas exposes a CMS-1500 box 22 resubmission code on every claim coverage (6 corrected, 7 replacement, 8 void), along with a Payer ICN field and a "Display CORRECTED CLAIM on HCFA header" checkbox. On a Claim.MD instance that code does real work. On a Candid instance it does nothing at all, and nothing tells you so.

A biller sets the code to 7, saves, re-queues the claim, gets a green result, and the claim files with the payer as an ordinary original claim. No error, no warning, no comment. The correction they think they filed does not exist.

## Why there's no passthrough fix

`POST /api/encounters/v4` has no claim-frequency, corrected-claim, or original-reference field anywhere in its 52 top-level fields. The only `claim_frequency_code` in the schema hangs off `external_claim_submission.submission_records[]`, which is for reporting claims filed *outside* Candid, the opposite of what this plugin does.

Candid confirmed this directly:

> The correct workflow here is to submit corrected claims in the Candid UI after the original claim is adjudicated (to avoid getting duplicate denials). There is no way to do this via API. Voids, corrected claims, and fresh claims all follow the same workflow in Candid. Candid populates the correct frequency code based on the type of submission selected.

So this isn't a gap we can close by wiring up a field. Corrections are a Candid UI workflow by design, and this PR makes the dead end visible instead of silent.

## What this does

Adds `resubmission_code_comment()` to `effect_helpers.py` and calls it from `CandidSubmitAPI.post()`. When any active coverage on the claim carries a resubmission code, the plugin appends a claim comment:

> Resubmission code 7 (replacement of prior claim) is set on this claim. Candid has no API field for it, so it was not sent and this claim was filed as an original. Corrections and voids are submitted in the Candid UI, where Candid sets the frequency code from the submission type you choose. Wait for the original claim to adjudicate before submitting the correction, otherwise it can come back as a duplicate denial.

The adjudication ordering is load-bearing, not padding. Correcting before the original adjudicates draws the exact duplicate denial the workflow exists to avoid, so the comment carries the ordering rather than just the destination. There's a test asserting on it so it can't quietly drop out of the string later.

The comment fires on success and on failure, because a biller expecting a correction needs to know either way.

**The payload is untouched.** This is additive only, so it cannot regress any existing submission.

## Why warn and not block

Blocking was the first instinct, and the data argued against it. Five instances running this plugin (four production, one sandbox) already have resubmission codes sitting on coverages. Those codes are inert today. If submission started failing whenever one was present, the next re-queue on those production instances would break. A comment can't do that to anyone.

Worth flagging for whoever picks this up: those production customers have the same silent problem, they just haven't noticed yet. Support is following up with them separately.

## What this deliberately doesn't do

It doesn't touch the resubmission controls in home-app. Hiding them on Candid instances is the obvious next thought and it's a trap: the same field drives box 22 on the printed CMS-1500, and a Candid customer can still legitimately print and mail a paper corrected claim. That trade-off needs a product decision, tracked separately.

## Testing

`uv run pytest tests/ -v` in `extensions/candid` gives 142 passed (135 existing + 7 new).

- 5 in `tests/test_effect_helpers.py` covering the helper: no code set, a single code, multiple coverages listed primary-first, coverages without codes ignored, and an unmapped code falling back to its raw value
- 2 in `tests/test_submit.py` covering the wiring: the comment lands in the returned effects with the payload unchanged, and an ordinary claim gets no extra comment

Verified non-vacuous by mutation. Neutering `resubmission_code_comment` to always return `None` fails 5 of the 7. The 2 that still pass are the negative tests, which is the correct result.

## Files

- `extensions/candid/candid/effect_helpers.py` - new `RESUBMISSION_CODE_LABELS` + `resubmission_code_comment()`
- `extensions/candid/candid/api/submit.py` - call it from `post()`
- `extensions/candid/candid/README.md` - document the limitation under Claim Submission
- `extensions/candid/tests/test_effect_helpers.py` - 5 tests
- `extensions/candid/tests/test_submit.py` - 2 tests

Generated from a Support Stack investigation.


bexport-gh:
cerberus review

bexport-gh:
cerberus review

PR #438, opened 2026-08-14
fix(external-calendar-busy-blocks): window-check RECURRENCE-ID overrides against their own times

bexport-gh:
## What

A recurring instance moved **earlier** in the day was re-created on the provider's Admin calendar every 15 minutes, until the base occurrence it replaced aged out. Observed on prod `ovadiahearthealth`: 16 duplicate `Busy` blocks in four hours on one provider's calendar (Pylon 32976).

## The bug

`_parse_base_event` discovers overrides by walking base occurrences, and it applied the window test to the **base occurrence's** time rather than the override's:

```python
if moment + duration <= now:   # base occurrence's time
    continue
...
override_event = _build_override_event(...)
if override_event is not None:
    out.append(override_event)  # appended unchecked
```

So an instance moved earlier could pass that test while its own moved slot had already ended. `live_busy_events` selects `ends_at > now`, so the ended block was invisible on the reconcile's "have" side while the unfiltered override kept it on the "want" side. Every tick concluded a block was missing and created one. Each new block was itself already ended, so the next tick couldn't see it either.

The loop runs from the moved slot's end to the base occurrence's original end. On ovadiahearthealth that was 17:00 to 21:00 UTC, which is exactly 16 fifteen-minute ticks. Moving an instance to an earlier *day* would cost ~96 blocks per day of displacement.

## The fix

- `ics/parser.py`: window-check the override against its own `starts_at`/`ends_at`. Drop the instance rather than falling back to the base time, since an override *replaces* the occurrence it names (RFC 5545 3.8.4.4) and emitting the base slot would block time the provider moved away from.
- `sync/cron.py`: backstop in `_reconcile` refusing to create any block whose `ends_at` has already passed, logged as a `WARNING`. Such a create is never useful and is unreachable by every later tick, so this closes the class rather than the single instance. Keeps the underlying parser bug findable instead of silently absorbing it.
- Bumps to `0.4.1`.

## This is not a #433 regression

Worth stating plainly, since #433 is the obvious suspect: `ics/parser.py` was not among the 13 files it touched. The defect is older.

The pre-0.4.0 cron hit the **same** window mismatch. It built its "have" side from `ImportedEvent.objects.filter(staff_id=..., ends_at__gte=now)`, so an ended override was excluded there too and it took its create branch every tick as well. It just couldn't complete the write: the create branch also inserted a tracking row, and `UniqueConstraint(staff_id, ics_uid, recurrence_id)` rejected it. The exception escaped `_diff_and_emit`, the per-feed handler caught it, and the tick's effects were discarded before reaching the calendar.

Confirmed in Elastic on the same feed nine days before #433 merged:

```
ERROR 2026-08-03 14:00:11  Unexpected error syncing feed 3; skipping
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint
  "canvas_medical__external_calendar_busy_blocks_importedevent_ext"
```

That error fired at 19:00:11 UTC; the block it corresponds to ends at exactly 19:00:00 UTC. Same end-of-block trigger, same feed.

So the tracking table was doing two jobs: the one it was designed for, and accidentally acting as an idempotency key that suppressed this bug. Making `ImportedEvent` inert was right for the orphan problem and removed that guard at the same time. The failure mode flipped from failing closed (feed silently skipped) to failing open (a duplicate block every 15 minutes).

## Tests

172 passing (was 167). Each guard was verified by removing it and confirming the new tests fail:

| Guard removed | Failing |
|---|---|
| parser override window check | 3 in `test_parser.py` |
| cron ended-create guard | 2 in `test_cron.py` |

`override_moved_earlier.ics` mirrors the production shape: a 5-hour Friday series with one instance moved 4 hours earlier. Tests assert it is yielded while in progress, gone once its own end passes, that it does not fall back to the base slot, and that the other Fridays in the window survive.

## Known limitation, not fixed here

Overrides are only reachable by walking base occurrences, and expansion starts at `now - base_duration`. An instance moved **later** across that boundary is therefore never discovered, so its block is never created. That is an under-block (provider looks more available than they are), the opposite direction from this bug. Fixing it properly means iterating overrides independently of base expansion, which is a real refactor and deliberately out of scope here.

## Also spotted while investigating

`BYSETPOS` is on the `UNSUPPORTED_PARTS` reject list (`ics/rrule.py:23`) and `parse_rrule` raising drops the **entire** VEVENT, so a provider's whole recurring commitment produces no blocks at all. One series on `ovadiahearthealth` is currently dropping on every tick. Filed separately; not addressed here.

_Prepared from a Support Stack investigation._


bexport-gh:
cerberus review

bexport-gh:
@alexiadowns-canvas (cc @rmagier1) — ready for review + merge when you have a moment. Flagging you since you merged #433, which is the code this touches.

**What it fixes:** a recurring instance moved *earlier* in the day was re-created on the provider's Admin calendar every 15 minutes. `_parse_base_event` discovers overrides by walking base occurrences and applied the window test to the base occurrence's time rather than the override's, so a moved instance could pass it while its own slot had already ended. `live_busy_events` selects `ends_at > now`, so the ended block was invisible on the reconcile's "have" side while the unfiltered override kept it on the "want" side. Observed on prod `ovadiahearthealth`: 16 duplicate blocks in four hours, which is exactly the number of ticks between the moved slot's end and the base occurrence's original end.

**The fix:** window-check the override against its own `starts_at`/`ends_at`, and drop the instance rather than falling back to the base time (an override replaces the occurrence it names, so emitting the base slot would block time the provider moved away from). Plus a backstop in `_reconcile` refusing to create any block whose `ends_at` has already passed, since such a create is never useful and is unreachable by later ticks.

**Worth knowing:** this is *not* a #433 regression. `ics/parser.py` wasn't among the files it touched. The pre-0.4.0 cron hit the same window mismatch, but its `ImportedEvent` insert violated the unique constraint on `(staff_id, ics_uid, recurrence_id)`, so the exception discarded the tick's effects and the create never reached the calendar. There's a `UniqueViolation` in Elastic on this same feed from Aug 3, nine days before #433 merged, firing at exactly a block's `ends_at`. Making `ImportedEvent` inert was right for the orphan problem and removed that accidental guard at the same time, which is what made the older bug visible. Details in the PR body.

**Evidence:**
- Cerberus: clean (0 important · 0 nit) on `d72aa7ab9`.
- 172 tests passing (was 167). Each guard verified by removing it and confirming the new tests fail: 3 in `test_parser.py`, 2 in `test_cron.py`.
- `override_moved_earlier.ics` mirrors the production shape (5-hour Friday series, one instance moved 4 hours earlier). Asserts it's yielded while in progress, gone once its own end passes, doesn't fall back to the base slot, and that the other Fridays survive.
- Transparency: I ran pytest but not the full `/wrap-up` (no mypy or branch-coverage run), so please don't read Cerberus's silence on those as a pass. It's a 22-line change, but your call whether you want those run first.

**Not fixed here, deliberately:** overrides are only reachable by walking base occurrences from `now - base_duration`, so an instance moved *later* across that boundary is never discovered and its block is never created. That's an under-block, the opposite direction, and fixing it means iterating overrides independently of base expansion. Separate refactor.

Bumps to `0.4.1`. Related: SUPPORT-2354 filed for a `BYSETPOS` gap in the same file that drops whole series (also an under-block).

_Prepared from a Support Stack investigation._

PR #439, opened 2026-08-17
Add POC Lab Test to Commit All Commands, rename protocols to handlers

rmagier1:
Refs [KOALA-6185](https://canvasmedical.atlassian.net/browse/KOALA-6185), which tracks keeping this button in step with the SDK.

## What should be added: `pocLabTest`, and only that

The right gate for "can a plugin commit this" turned out not to be the SDK — `BaseCommand.commit()` will happily synthesize a `COMMIT_<KEY>_COMMAND` effect for any command — but whether home-app registers an interpreter for that effect in `plugin_io/interpreters/commands/__init__.py`. There are **34** active registrations. Against the button's 31 mappings, that leaves three, and only one belongs here:

| Command | Verdict | Why |
| --- | --- | --- |
| `pocLabTest` | **added** | `CommitPOCLabTestCommand(CommitCommandEffectInterpreter): pass` — plain generic interpreter, and the command constructs and commits with only a `command_uuid`, so no `extra_params` handling needed |
| `labOrder` | skipped | committable, but an order — it has to be sent, not merely committed |
| `reference` | skipped | committed on origination, so it never sits `staged` and the mapping would be dead code |

`reasonForVisit` stays staged by design and is never committed. Consistent with that, its interpreter registration is **commented out** (`plugin_io/interpreters/commands/__init__.py:452`), so the effect would not be honored.

The ordering and prescribing commands — `prescribe`, `refill`, `adjustPrescription`, `imagingOrder` — have `COMMIT` effects in the SDK but no interpreter at all, which is the mechanism behind "can't be committed". `refer` carries a delegate, so a staged Refer doesn't imply the provider intended to commit it.

## Requires an SDK bump

`uv.lock` was pinned to **canvas 0.82.0**, which predates `POCLabTestCommand` — importing it fails outright:

```
ImportError: cannot import name 'POCLabTestCommand' from 'canvas_sdk.commands'
```

`POCLabTestCommand` first shipped in **0.162.0**. The lock moves to 0.198.0 and `CANVAS_MANIFEST.json` declares 0.162.0 as the floor (it previously declared 0.44.3).

## protocols → handlers

Renamed throughout, matching the SDK's preferred terminology: the package directory, the manifest `components` key and class path, the test package, and the "A protocol that adds…" descriptions in the manifest and `pyproject.toml`. Git tracked all four file moves as renames.

## README now explains the omissions

Previously the exclusions weren't written down anywhere, so the button looked like it just hadn't caught up. The README now groups every omitted command by reason — orders needing to be sent, Refer's delegate, RFV always staying staged, Reference being originate-committed, Reviewed/Custom Command having no COMMIT effect, and commands with no SDK equivalent.

## Fixes two stale counts

Both the mapping test and the README claimed **27** command types against a map of **31** — neither was updated when the four review commands were added, so `test_all_command_types_mapped` has been failing on `main` since then.

The test now asserts an explicit set of schema keys rather than a count: a count can't say *which* command is missing when it fails, and it rots silently. The README's supported list is checked against the handler's map (32/32).

## Verification

`16 passed` — including the mapping test that was red before this branch.

## Note for a follow-up, not fixed here

`self.context` is deprecated as of SDK 0.11.0 in favour of `event.context` (`handlers/commit_all_commands.py:97`), and is slated for removal in 1.0.0. Left alone to keep this PR scoped.

[KOALA-6185]: https://canvasmedical.atlassian.net/browse/KOALA-6185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rmagier1:
Pushed d9fa498 — makes the button dynamic, per the [reload action buttons effect](https://docs.canvasmedical.com/sdk/effect-reload-action-buttons/).

**`visible()` now requires something to commit.** Previously it rendered on every unlocked note. It now also needs at least one staged command whose schema key is in `SCHEMA_KEYS_TO_COMMANDS` — a single `.exists()` query, extracted alongside `handle()`'s lookup as `staged_commands()`.

Scoping to the *mapped* keys is the part that matters. A note holding only commands this button can't commit — an unsent Prescribe, say — would otherwise show a button that does nothing when clicked.

**`handle()` appends a `ReloadNoteActionButtonsEffect`.** Committing empties the staged set, so the visibility condition stops holding; the reload makes the button disappear immediately rather than lingering until the next page load.

Two details worth calling out:

- **It's appended last.** Effects apply in order, so the reload has to observe the commits — otherwise it re-evaluates visibility against the pre-commit state.
- **It's skipped when nothing committed** (nothing staged, or every command failed validation). Visibility can't have changed, so the round trip would be wasted.

**Two constraints I hit:**

- The effect takes the note's **UUID**, but `context["note_id"]` is a **dbid** — the SDK's models use `dbid` as their primary key (`canvas_sdk/v1/data/base.py:87`). Hence the `Note.objects.get(dbid=...)` lookup.
- `ReloadNoteActionButtonsEffect` landed in **canvas 0.179.0**, later than the 0.162.0 floor this PR previously set for `POCLabTestCommand`, so the manifest floor moves again. It's also not re-exported from `canvas_sdk.effects`, so it's imported from `canvas_sdk.effects.action_button`.

I also verified `from django.db.models.query import QuerySet` (used for the extracted method's return type) is sandbox-allowed — `plugin_runner/sandbox.py:480-483`.

**Tests: 19 passed.** An autouse fixture stubs the note lookup and reload effect, since every committing test now carries the trailing effect. Four new tests: two for the visibility rule (including the "only unmappable commands staged" case), two for the reload — that it's appended last exactly once, and that it's absent when nothing committed.

README gained a **When the Button Appears** section covering both conditions and the reload behaviour.

PR #440, opened 2026-08-27
Fix PayTheory card form blank on re-open by serving it from a same-origin iframe

alexiadowns-canvas:
## Summary

Fix PayTheory card form failing on iframe-based rendering by adding a submit bridge between the host DOM and the iframe, and breaking a visibility deadlock.

**Problem:** After moving the PayTheory card form into an iframe to fix the custom element re-registration crash (KOALA-5882), two issues remained:

1. **No submit button in host DOM.** Canvas's `CustomPayment.tsx` finds the submit trigger via `querySelector('#submit')` in the host document. With the form inside an iframe, that element was gone — clicking "Pay" or "Save Card" did nothing.
2. **Hidden-until-ready deadlock.** Canvas hides the payment form until `setFormIsReady(true)` is called, but PayTheory's `readyObserver` only fires when the hosted fields become visible. Neither side yields first.

**Fix (this PR):**

- **Submit bridge:** A hidden `<button id="submit">` in the host DOM forwards clicks into the iframe via `postMessage({ type: "pt-submit" })`. Inside the iframe, a `message` event listener calls the extracted `doTokenize()` function. The in-iframe `#submit` click listener is preserved for standalone use.
- **Unconditional `setFormIsReady(true)`:** Called immediately after `paytheory.payTheoryFields()` is invoked, in addition to the existing `readyObserver` callback. This breaks the deadlock so the form becomes visible and PayTheory can finish initializing.

Refs: KOALA-5882, KOALA-5881

## Changed files

- `handlers/paytheory_payment_processor.py` — `_card_fields_html()` now returns iframe + hidden #submit + postMessage bridge script
- `templates/card_form.html` — extracted `doTokenize()`, added `message` listener for `pt-submit`, added unconditional `setFormIsReady(true)` call
- `tests/handlers/test_paytheory_payment_processor.py` — new `TestCardFieldsHtml` class asserting iframe src, #submit presence, and postMessage bridge; updated `TestPaymentForm` and `TestAddCardForm` to match new HTML output

## Test plan

- [ ] Open Collect Payment modal, verify card fields render
- [ ] Enter card details and submit payment — confirm tokenization completes
- [ ] Close and re-open the modal — confirm no "custom element already defined" error
- [ ] Add a saved card via "Add Card" flow — confirm it persists

### DIFF
--- a/extensions/paytheory-payment-processor/paytheory_payment_processor/CANVAS_MANIFEST.json
+++ b/extensions/paytheory-payment-processor/paytheory_payment_processor/CANVAS_MANIFEST.json
@@ -51,6 +51,10 @@
                     "read": [],
                     "write": []
                 }
+            },
+            {
+                "class": "paytheory_payment_processor.handlers.card_form_api:CardFormAPI",
+                "description": "Serves the PayTheory hosted-fields card form as a standalone HTML page, loaded in an iframe to avoid custom element re-registration errors (KOALA-5882)."
             }
         ],
         "commands": [],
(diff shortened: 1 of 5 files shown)


alexiadowns-canvas:
cerberus review

PR #443, opened 2026-08-31
Fix PayTheory card save — pass payorId during tokenization so cards l…

alexiadowns-canvas:
…ink to the patient

The iframe card form was calling tokenizePaymentMethod({}) with no payor association. Cards were tokenized but never linked to a payor, so get_payment_methods always returned 0 results. Now payment_form() and add_card_form() resolve the patient's payor_id and pass it through the iframe URL to the template, which passes it to tokenizePaymentMethod().


alexiadowns-canvas:
cerberus review

PR #449, opened 2026-09-10
fix(schedule_view): don't offer Cancel/No Show on schedule events

bexport-gh:
The appointment modal renders Cancel and No Show on rows the API rejects. Both endpoints guard on `if not appt.note_id` (`handlers/schedule_api.py`), and schedule events — room reservations, generic admin events — never have a note. Clicking either button returns 400 "Appointment has no associated note and cannot be cancelled."

This surfaced on a downstream fork of this extension, where staff hit the error repeatedly trying to cancel from the room row rather than the visit row. The reference has the same defect, and slightly worse: there's no branch separating these rows at all, so both buttons render unconditionally.

The `note_id` guard itself is correct and stays. It exists because the endpoint used to return 200 before the async SDK effect ran, so failures were silent and the card stayed on screen. It turned a silent no-op into a visible error. This change removes the button that can only ever produce it.

Gated on `!appt.note_id` rather than `is_schedule_event || is_block`. `note_id` serializes to `""` when null, so the JS condition is exactly the server's rejection condition and can't drift. The flag version would: a room booking whose provider lost the `RR` role computes `is_schedule_event = false` and slips through to a guaranteed 400.

Cancelling the visit that owns a room already releases the room via the `scheduling_with_rooms` `APPOINTMENT_CANCELED` cascade, so the replacement note points there rather than leaving the section empty.

No test suite covers this extension. The change is a render-path guard with no state change.


bexport-gh:
cerberus review

PR #450, opened 2026-09-15
gcal_sync: expand multi-day all-day Google events into one hold per day

bexport-gh:
Fixes [PLUGIN-445](https://canvasmedical.atlassian.net/browse/PLUGIN-445).

## The defect

Canvas blocks a hold's **date of service** and nothing after it. `gcal_sync` imported every inbound Google event as exactly one hold, so a multi-day all-day PTO event blocked only its first day.

Google reports an all-day event's `end.date` exclusively, so PTO entered for 8/31 through 9/4 arrives as start 8/31 / end 9/5. `parse_event_window` turned that into a single 7200-minute window, `build_hold_effect` emitted one `ScheduleEvent.create()`, and 9/1 through 9/4 stayed bookable for that provider.

It is a silent logic bug. The hold is created successfully, so there is no Sentry error and nothing in the logs to notice.

## The fix

An all-day event expands into one hold per covered day across `[start.date, end.date)`, each 1440 minutes, each stamped `{google_event_id}:{YYYY-MM-DD}` in `external_identifiers`. The covered set is reconciled on every pull by `InboundSync._reconcile_all_day_holds`, so extending, shortening, moving, or deleting the Google event adds and removes the matching per-day holds in one pass.

Timed events keep the existing single-hold path and the bare event id, unchanged.

### Edges worth a reviewer's attention

- **A pre-existing bare-id hold is adopted as day one**, not duplicated. This is what lets an already-imported event gain its missing days instead of minting a second hold for day one, and it is the path every currently-affected instance takes on its next sync.
- **The no-op guard now requires the day set to be complete**, not just a matching content hash. A hold written by the old behavior matches on hash, so without the completeness check its missing days would never be created.
- **The convergence guard runs per day** (`_external_value_exists`), so a day whose hold was cancelled out-of-band stays cancelled and the create/cancel/re-create loop stays closed at per-day granularity.
- **The in-flight marker gates only creates.** Updates and removals are idempotent, so a partially created set finishes itself once the marker ages past the grace window.
- **Cancellation removes every hold stamped with the event id.** A cancellation delta carries no start/end, so the covered dates cannot be recomputed; `_hold_delete_effect` becomes `_hold_delete_effects` and goes by the stamped ids.
- **Expansion is capped at 186 days**, sized just past the +6-month inbound import window so the window stays the binding constraint. Real events run long: one live install carries a 134-day all-day hold.

`_live_holds_for_event` uses a `value__startswith` DB filter and then an exact separator check in Python, because a prefix match alone also catches a different event whose id merely starts with this one. Google event ids never contain `:`.

## Fleet state (measured 2026-09-15)

Four production instances carry multi-day `gcal-sync` holds:

| Instance | Multi-day holds | Live | Live + future | Longest |
|---|---|---|---|---|
| riviamind | 592 | 58 | 13 | 90,840 min (63 d) |
| thelanby | 8 | 7 | 2 | 193,020 min (134 d) |
| decipherhealth | 7 | 2 | 0 | 4,380 min (3 d) |
| rebel | 3 | 3 | 0 | 14,400 min (10 d) |

Provider time currently open despite PTO: riviamind 13 holds / 13 providers / 100 provider-days (2026-10-12 to 2026-12-29), thelanby 2 holds / 2 providers / 141 provider-days (2026-09-21 to 2026-10-27).

Riviamind's newest `gcal-sync` identifier is stamped 2026-09-15, so these are still being written.

```sql
SELECT count(*) AS multiday_holds,
       count(*) FILTER (WHERE a.status <> 'cancelled') AS live_multiday,
       count(*) FILTER (WHERE a.status <> 'cancelled' AND a.start_time > now()) AS live_future,
       max(a.duration_minutes) AS max_minutes
FROM api_appointment a
JOIN api_appointmentexternalidentifier x ON x.appointment_id = a.id
WHERE x.system = 'gcal-sync' AND a.duration_minutes > 1440
HAVING count(*) > 0;
```

`nextgenmed` and `salicefamilyhealth` returned connection errors and contributed no data to that sweep.

## Tests

375 passing, up from 343. New file `tests/test_inbound_all_day.py` covers date expansion and the exclusive `end.date`, the per-day identifier and 1440-minute window, adoption of a pre-existing bare-id hold, stale-day removal when the event shortens, the completeness check on an unchanged hash, both ingest toggles, the in-flight and per-day convergence guards, `force_rebuild`, and removal of every per-day hold on cancellation. Coverage 99% on both changed modules; mypy clean on both.

## Follow-ups this does not carry

- Existing wrong holds re-block only on a re-sync per provider (admin "Re-import") or a manual block. That is an operational step per affected instance.
- thelanby, decipherhealth, and rebel run their own copies of the plugin and need this landed there or an install of the updated reference. riviamind's port is a companion PR on `canvas-medical/gtm-extensions`.
- `_within_import_window` tests only an event's start, so a long PTO that began over a month ago but runs into the future is skipped entirely. Pre-existing and separate from this defect.

Generated from a Support Stack investigation.

[PLUGIN-445]: https://canvasmedical.atlassian.net/browse/PLUGIN-445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bexport-gh:
cerberus review

